PR #5, opened 2018-07-06
Add documents to CreateCase request

nickmcmahonHOD:


zuzak:
Looks like I've written too many ways to create a case with documents here?

nickmcmahonHOD:
lgtm

PR #152, opened 2019-04-09
Cache info resources.

nickmcmahon01:


src/main/java/uk/gov/digital/ho/hocs/workflow/application/CacheWarmer.java
```diff
@@ -0,0 +1,85 @@
...
+            Set<CaseDataType> caseDataTypeSet = this.infoClient.getCaseTypes();
+            for(CaseDataType caseDataType : caseDataTypeSet) {
+                this.infoClient.populateCaseTypeByShortCode(caseDataType.getShortCode(), caseDataType);
+                Set<SchemaDto> schemas = this.infoClient.populateSchemasForCaseType(caseDataType.getDisplayCode());
+                for(SchemaDto schemaDto : schemas) {
+                    this.infoClient.populateSchema(schemaDto.getType(), schemaDto);
+                }
+            }
+        } catch(Exception e) {
+            log.warn("Failed to prime Case Types. : {}",e.toString(), value(EVENT, CACHE_PRIME_FAILED));
```
SWhitlam: Same as the other review, set the exception value rather than appending it to the log message.

PR #192, opened 2019-10-01
HOCS-1069

Paulcr-ho:
Adding the correct logging for the migration data.
Added a default value setting on the document exception.


hubert-raich:
Should we close this PR? Some login changes might be useful in the future but logic changes are not needed at this point

PR #239, opened 2020-01-13
HOCS-1408 - Record eligibility rejection reason

hubert-raich:


hubert-raich:
I've created a step 'Clear Eligibility Rejection Reason' that clears the variable in camunda and db

MarkWrightBjss:
> I've created a step 'Clear Eligibility Rejection Reason' that clears the variable in camunda and db

Yep, see that now. I was looking at the "BACKWARD" path. If the user selects a rejection value, but then clicks back, would this save the rejection value and bypass the clearing?

hubert-raich:
Confirmed with Mark that 'back' action does not save any of the fields on the screen so rejection reason would not be saved.

PR #245, opened 2020-01-21
Feature/wcs

Paulcr-ho:


Paulcr-ho:
There are issues with the drone build pull of the merge.

PR #290, opened 2020-05-22
HOCS-1739: Modify WorkflowService to use modified info-service  endpo…

parth-ho:
…int and new caswork endpoint.

### DIFF
--- a/src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
+++ b/src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
@@ -9,6 +9,7 @@ import uk.gov.digital.ho.hocs.workflow.api.dto.*;
 import uk.gov.digital.ho.hocs.workflow.client.camundaclient.CamundaClient;
 import uk.gov.digital.ho.hocs.workflow.client.caseworkclient.CaseworkClient;
 import uk.gov.digital.ho.hocs.workflow.client.caseworkclient.dto.CreateCaseworkCaseResponse;
+import uk.gov.digital.ho.hocs.workflow.client.caseworkclient.dto.GetAllStagesForCaseResponse;
 import uk.gov.digital.ho.hocs.workflow.client.caseworkclient.dto.GetCaseworkCaseDataResponse;
 import uk.gov.digital.ho.hocs.workflow.client.documentclient.DocumentClient;
 import uk.gov.digital.ho.hocs.workflow.client.infoclient.InfoClient;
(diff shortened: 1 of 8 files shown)


src/test/java/uk/gov/digital/ho/hocs/workflow/client/caseworkclient/CaseworkClientTest.java
```diff
@@ -150,4 +147,22 @@ public void updateDataValue() {
         verify(restHelper).put(eq(caseServiceUrl), eq(expectedUrl), eq(newValue), eq(Void.class));
         verifyNoMoreInteractions(restHelper);
     }
+
+    @Test
+    public void GetAllStagesForCase() {
```
hubert-raich, line 152: make the method name start with lowercase letter (getAll)

src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
```diff
@@ -131,10 +153,17 @@ public GetCaseDetailsResponse getReadOnlyCaseDetails(UUID caseUUID) {
         GetCaseworkCaseDataResponse inputResponse = caseworkClient.getFullCase(caseUUID);
```
hubert-raich, line 153: I feel like this call could have been pushed up and made to contain the list of 'visited' stages (in order). Especially with the 'full' flag being passed in which we could use to drive the logic around when should we include the list of stage names.

src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
```diff
@@ -114,11 +116,31 @@ public GetCaseResponse getAllCaseStages(UUID caseUUID) {
 
         GetCaseworkCaseDataResponse inputResponse = caseworkClient.getFullCase(caseUUID);
 
-        Set<SchemaDto> schemaDtos = infoClient.getSchemasForCaseType(inputResponse.getType());
+        GetAllStagesForCaseResponse allStagesForCase = caseworkClient.getAllStagesForCase(caseUUID);
```
hubert-raich, line 119: An 'ok' solution, I think it would be better for performance if we only fetched from 'casework' once. Pulling case data + list of visited stages in one call (more comments underneath).  As SchemaDto already has stageType property we could then filter it out here without making any changes in info service for schema retrieval. 
Current setup forces the case data to be read twice especially with no dedicated method in casework. To make it a little bit quicker we could have created a new method in casework that only fetched list of stageTypes the case has been through.

hubert-raich: Just to make things clear, this is just advice on how we could have done it slightly neater, no need to action this item at the moment:)

src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
hubert-raich: Just to make things clear, this is just advice on how we could have done it slightly neater, no need to action this item at the moment:)

PR #312, opened 2020-06-25
HOCS-1812 - WCS case closed with no apparent reason

hubert-raich:
- updating default id generation (DBIdGenerator) in favor of StrongUuidGenerator
as it is the one recommended in camunda docs
- Adding java-uuid-generator dependency as it is required for StrongUuidGenerator

After investigation is has been discovered that intermittent issues are caused
by optimistic lock exception that occur when Camunda uses default DbIdGenerator on
more than 1 service (single db and multiple services)

More details here:
https://docs.camunda.org/manual/7.9/user-guide/process-engine/id-generator/#the-uuid-generator

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -66,6 +66,7 @@ dependencies {
 	implementation('net.logstash.logback:logstash-logback-encoder:5.2')
 
 	implementation('org.camunda.bpm:camunda-engine-spring:7.9.0')
+	implementation('com.fasterxml.uuid:java-uuid-generator:3.1.0')
 
 	runtimeOnly('org.postgresql:postgresql')
 	compileOnly group: 'org.projectlombok', name: 'lombok', version: '1.18.6'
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -66,6 +66,7 @@ dependencies {
 	implementation('net.logstash.logback:logstash-logback-encoder:5.2')
 
 	implementation('org.camunda.bpm:camunda-engine-spring:7.9.0')
+	implementation('com.fasterxml.uuid:java-uuid-generator:3.1.0')
```
zuzak, line 69: There's a 4.0.1, a 3.2.0 and a 3.3.0 in Maven, are we deliberately using an older version?

hubert-raich: Yep, 3.1 was listed as this is what camunda 7.9 is using. I've updated the version to 3.1.2 to match exactly what Camunda 7.9 spec says.

PR #317, opened 2020-07-07
HOCS-1824: Add stage deadlines for Home Secretary sign off cases.

parth-ho:


parth-ho:
Yeah was thinking about doing that but couldn't immediately see how to pass a non-string arg with camunda. `updateValue` seems like another way to do it though.

PR #336, opened 2020-10-16
Feature/hocs 1921 converted case notes mandatory

abarker66:
Additions for mandatory case notes when converting a case from ministerial to official or vice versa.


LiamMacP (approved):
Changes pulled, and checked. Camunda files look good. 

Task to be created to pull out the case conversion to a subtask, should be able to use variables to remove added Bpmn method. 

Approved

PR #337, opened 2020-10-22
HOCS-2025: remove stage for saving notes when deleting cases

LiamMacP:
When deleting cases that are duplicates after escalating, we no longer want to capture the reason for this.


abarker66 (approved):
Camunda updates look good.

PR #342, opened 2020-11-27
HOCS-2191 Don't assign team from topic if inactive

zuzak:
This commit adds a check immediately prior to assigning a team based on
a selected topic at (e.g.) markup stage. If the team is inactive, it
doesn't assign a team at all.

As far as I can tell, that means the case gets sent to the default team
(that is, DCU's central team).

This has the side-effect of meaning that in the UI on the override
("Answering") screen the missing team will show up as an empty field.
However, this is quite an edgey edge-case that should rarely surface, so
I think that's fine for now.

This commit also adds tests; previously the method had no coverage.

### DIFF
--- a/src/main/java/uk/gov/digital/ho/hocs/workflow/BpmnService.java
+++ b/src/main/java/uk/gov/digital/ho/hocs/workflow/BpmnService.java
@@ -144,8 +144,12 @@ public class BpmnService {
 
         Map<String, String> teamsForTopic = new HashMap<>();
         TeamDto teamDto = infoClient.getTeamForTopicAndStage(caseUUID, topicUUID, stageType);
-        teamsForTopic.put(teamNameKey, teamDto.getUuid().toString());
-        teamsForTopic.put(teamUUIDKey, teamDto.getDisplayName());
+        if (teamDto.isActive()) {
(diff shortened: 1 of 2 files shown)


MarkWrightBjss (approved):
Good work. Definitely better than assigning to an inactive team.

LiamMacP:
Merging so it can get in the next QA release.

src/main/java/uk/gov/digital/ho/hocs/workflow/BpmnService.java
```diff
@@ -144,8 +144,12 @@ public void updateTeamsForPrimaryTopic(String caseUUIDString, String stageUUIDSt
 
         Map<String, String> teamsForTopic = new HashMap<>();
         TeamDto teamDto = infoClient.getTeamForTopicAndStage(caseUUID, topicUUID, stageType);
-        teamsForTopic.put(teamNameKey, teamDto.getUuid().toString());
-        teamsForTopic.put(teamUUIDKey, teamDto.getDisplayName());
+        if (teamDto.isActive()) {
```
MarkWrightBjss, line 147: Is there any chance that 'teamDto' will be null at this point? (e.g. if topic has no assigned team)

PR #344, opened 2020-12-03
HOCS-2019 : correct bpmn calls to bpmn service to create case note

abarker66:
Corrections made to the calls made from bpmn diagrams to the bpmn Service method to create case notes.
The call to createCaseNote from bpmn diagrams had the wrong parameters, now fixed.

### DIFF
--- a/src/main/resources/processes/MPAM_DISPATCH.bpmn
+++ b/src/main/resources/processes/MPAM_DISPATCH.bpmn
@@ -206,7 +206,7 @@
       <bpmn:outgoing>Flow_13k7b3x</bpmn:outgoing>
       <bpmn:outgoing>Flow_05292wb</bpmn:outgoing>
     </bpmn:exclusiveGateway>
-    <bpmn:serviceTask id="Activity_08te8ix" name="Save Close Case Notes" camunda:expression="${bpmnService.createCaseNote(execution.getVariable(&#34;CaseUUID&#34;),execution.processBusinessKey, execution.getVariable(&#34;CaseNote_CaseClose&#34;))}">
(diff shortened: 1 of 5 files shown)


MarkWrightBjss (changes requested):
1. Please add a unit test for the new "createCaseNote" endpoint.
2. Please add new caste note type to the frontend project.

abarker66:
The BpmnServiceTest mocks the caseworkClient so does not actually add stuff to the database so I don't see the point unless we change the philosophy of testing anything that adds stuff to the databases.I can add one but it will be as useless as the other tests in this class. All they really do is parameter type checking, but it is so mocked it does very little else.

abarker66:
This is blocked for now, the bpmn service tests are mocking all the clients used by bpmn service so are not really performing any functional tests at all. What value do they have??

abarker66:
I cannot mock the calls necessary to add a test for the bpmn.createCaseNote method, they have no return params. The only useful test for this and many other methods in this package would be full on functional checks to the database which would mean quite a bit of work but maybe worth doing for future tests because the existing tests are a nonsense really they are fully mocked and just rigged to pass.

abarker66:
I added a test for createCaseNote to prove that Mockito works. I think it would be well worth investing some time into writing proper functional tests for a focussed regression suite, we have the REST helper so we could call the real methods and read back real results from the database rather than mocking all the clients used by bpmnService and pretending we are testing it.

LiamMacP (approved):
Changes look good, tests added.

MarkWrightBjss (approved):

PR #345, opened 2020-12-14
HOCS-2136: Transfer to OGD at MPAM Creation stage

parth-ho:
This PR modifies/creates the camunda flows necessary for the transfer case flow.


abarker66 (changes requested):
MPAM_CREATION.bpmn looks like it is removing the MP Mandatory section of the diagram - is that intentional or is it a merge issue?

parth-ho:
@abarker66 yes just a conflict fix issue, should be all sorted now.

PR #347, opened 2020-12-21
Refactor/hocs 2275 separate integration tests

philbarton:
Some of the tests in the project have the dependency  of a running postgres database. This change marks those as integration tests, by moving them to an integration test directory and applying standard gradle config.
This allows unit tests to be run much quicker, and without the external dependency.

This change is a precursor to introducing more unit tests to the workflow project.


README.md
```diff
@@ -77,7 +77,10 @@ and there will be a new instance of postgres.
 ## Tests
 
 The suite of tests includes unit tests for the resource and services classes, and integration tests. In order to run the integration tests, an instance of postgres must be running.
-
+To run the unit tests in isolation (without the need for a postgres instance running), run the following :
```
zuzak, line 80:
```suggestion
To run the unit tests in isolation (without the need for a postgres instance running), run the following:
```

README.md
```diff
@@ -77,7 +77,10 @@ and there will be a new instance of postgres.
...
+```
```
zuzak, line 81:
```suggestion
```bash
```

build.gradle
```diff
@@ -68,11 +68,36 @@ dependencies {
...
+
+task integrationTest(type: Test) {
+	description = 'Runs integration tests.'
+	group = 'verification'
+	testClassesDirs = sourceSets.integrationTest.output.classesDirs
+	classpath = sourceSets.integrationTest.runtimeClasspath
+	shouldRunAfter test
+}
+
+check.dependsOn integrationTest
```
zuzak, line 103: This isn't a showstoppper, but ideally you'd have a new-line here so it plays better with Git.

(If you were to add a line 104, you'd need to modify line 103 to add the line ending, so `git diff` will be slightly strange)

philbarton: I like it, makes sense, good shout.

zuzak:
Your commit message is a bit strange.
The first line should be a summary that stands alone without further context.

If you're going to use (the optional) extra lines to add more context, the second line should always be blank, then you continue typing on line three and beyond.

Here you've put the Jira ticket on line 1, and then the actual message on line 2.
You can change it with 

```bash
git commit --amend -m "HOCS-2275 Separate integration tests" &&
git push origin +refactor/HOCS-2275-Separate_integration_tests # the plus means "force push but only to this branch"
```

These might sound like particularly persnickety comments, and to an extent they are -- but adhering to [these Git commit guidelines](https://chris.beams.io/posts/git-commit/) generally makes Git easier to work with. Squash these commits prior to merge for bonus points.

(I'm too bad at java to actually review this PR properly.)

Paulcr-ho (approved):

PR #348, opened 2021-01-05
Spike/hocs 2284 workflow process unit testing

philbarton:
This change adds an example of unit testing camunda process flows. It is a spike and not extensive. The 3 tests added should help as examples for more richer tests against future real stories.

The work is based on these notes : 

- https://camunda.com/best-practices/testing-process-definitions/
- https://camunda.com/blog/2020/10/testing-entire-process-paths/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 .gradle
 /build/
+/target/
 !gradle/wrapper/gradle-wrapper.jar
 
 ### STS ###
@@ -20,7 +21,6 @@
 
 ### NetBeans ###
 /nbproject/private/
-/build/
 /nbbuild/
 /dist/
 /nbdist/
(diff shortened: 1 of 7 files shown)


willhiorns:
Was thinking about this yesterday while Slack down.  Not expecting changes to PR, but raising for discussion.

I think that it is probably a code smell if a test needs to know about a String with value “CallActivity_1c4zy60”
 
Tests, even unit tests, should not know about internals. They are bone fide clients of the code under test. Coupling them to implementation detail leads to trouble. 
 
I think that this String is being used as a proxy for the thing we really care about. I also suspect that extracting it to a constant with a better name would just be smoke and mirrors concealing that fact, and would still leave us coupled (logically, by duplication of the value) to a private String that is not part of our application’s API (as in application programming interface).
 
So, what are we really trying to do when we assert that the Runnable process with that ID has completed? What is the outcome that should result?
 
 If it is client-visible state, assert on that returned value.
 If it is an interaction, assert on a mock.
 If it is a change in system behaviour, assert on that side-effect (ewww).

MarkWrightBjss:
New tests ran and passed locally. I have seen the future.

MarkWrightBjss (approved):
Even the build server seems to be happy.

PR #352, opened 2021-01-06
HOCS-2296 : change to Official clears Ministerial data, fix

MarkWrightBjss:
Changing a case from reftype "Ministerial" to "Official" clears ministerial only data values: MinSignOffTeam, Addressee
At both Triage and Draft.
Now with Camunda unit tests!

### DIFF
--- a/src/test/java/uk/gov/digital/ho/hocs/workflow/processes/MPAMDraft.java
+++ b/src/test/java/uk/gov/digital/ho/hocs/workflow/processes/MPAMDraft.java
@@ -1,0 +1,98 @@
+package uk.gov.digital.ho.hocs.workflow.processes;
+
+import org.camunda.bpm.engine.test.Deployment;
+import org.camunda.bpm.engine.test.ProcessEngineRule;
+import org.camunda.bpm.engine.test.mock.Mocks;
+import org.camunda.bpm.extension.process_test_coverage.junit.rules.TestCoverageProcessEngineRule;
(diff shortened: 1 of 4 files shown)


src/test/java/uk/gov/digital/ho/hocs/workflow/processes/MPAMDraft.java
```diff
@@ -0,0 +1,98 @@
...
+
+    @Before
+    public void setup() {
+        Mocks.register("bpmnService", bpmnService);
+    }
+
+    @Test
+    public void whenMinisterialChangedToOfficial_thenMinisterialValuesAreNotCleared() {
+
+        when(processScenario.waitsAtUserTask("UserTask_00s8k9f"))
```
LiamMacP, line 49: I don't mind if this gets done in this one, but would be nice to proactively go forward change the name of the tasks to help quicker identify what's been tested.

LiamMacP (approved):

src/test/java/uk/gov/digital/ho/hocs/workflow/processes/MPAMDraft.java
MarkWrightBjss: Okay, yeah, let's start doing this. It will make the code more readable.

philbarton:
Just noticed the use of the nasty label naming from Camunda. In my tests I'm creating constants to make them clearer.

PR #354, opened 2021-01-08
HOCS-2026 : new workstack "Rejected" column

MarkWrightBjss:
Set new "Rejected" value when moving back from: Draft, QA (to Triage), QA (to Draft), PO, Dispatch.
Clear new "Rejected" value when moving from: Triage->Draft, Draft->QA, Draft->Dispatch (B-Ref only).
Renamed some Camunda boxes to improve readability of workflow unit tests for each of the eight scenarios above.


Paulcr-ho (approved):
Nice! New testing process in place.

PR #355, opened 2021-01-08
HOCS-2266 : Return to private office stage from minister sign off.

philbarton:
Add a rejection step similar to "Not approved" but in this case go back to the PO sign-off stage.

### DIFF
--- a/src/test/java/uk/gov/digital/ho/hocs/workflow/processes/DCU_MIN.java
+++ b/src/test/java/uk/gov/digital/ho/hocs/workflow/processes/DCU_MIN.java
@@ -1,6 +1,5 @@
 package uk.gov.digital.ho.hocs.workflow.processes;
 
-import org.camunda.bpm.engine.delegate.DelegateExecution;
 import org.camunda.bpm.engine.test.Deployment;
 import org.camunda.bpm.engine.test.mock.Mocks;
 import org.camunda.bpm.extension.mockito.ProcessExpressions;
@@ -16,10 +15,13 @@ import org.junit.runner.RunWith;
 import org.mockito.Mock;
 import org.mockito.junit.MockitoJUnitRunner;
 import uk.gov.digital.ho.hocs.workflow.BpmnService;
+import uk.gov.digital.ho.hocs.workflow.util.ExecutionVariableSequence;
+import uk.gov.digital.ho.hocs.workflow.util.CallActivityReturnVariable;
 
-import java.util.function.Consumer;
+import java.util.Arrays;
 
-import static org.mockito.Mockito.*;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
 
 @RunWith(MockitoJUnitRunner.class)
 @Deployment(resources = {
@@ -35,12 +37,9 @@ import static org.mockito.Mockito.*;
         "processes/STAGE_WITH_USER.bpmn"})
 public class DCU_MIN {
 
-    public static final String APPROVE_MINISTER_SIGN_OFF = "UserTask_0eagz4p";
-    public static final String MIN_SIGN_OFF = "CallActivity_1syv7pu";
     public static final String DISPATCH = "CallActivity_1rowgu5";
     public static final String PO_SIGN_OFF = "CallActivity_0pmeblj";
     public static final String INITIAL_DRAFT = "CallActivity_1ket68y";
-    public static final String APPROVE_MIN_SIGN_OFF = "ServiceTask_0te5zh0";
     @Rule
     @ClassRule
     public static TestCoverageProcessEngineRule rule = TestCoverageProcessEngineRuleBuilder.create()
@@ -50,8 +49,6 @@ public class DCU_MIN {
     BpmnService bpmnService;
     @Mock
     private ProcessScenario dcuMinProcess;
-    @Mock
-    private ProcessScenario minSignOffProcess;
 
     @Before
     public void defaultScenario() {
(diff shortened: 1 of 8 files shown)


src/test/java/uk/gov/digital/ho/hocs/workflow/processes/DCU_MIN.java
```diff
@@ -0,0 +1,164 @@
...
+        "processes/DCU_MIN_MINISTER_SIGN_OFF.bpmn",
+        "processes/DCU_MIN_PRIVATE_OFFICE.bpmn",
+        "processes/DCU_BASE_DISPATCH.bpmn",
+        "processes/STAGE.bpmn",
+        "processes/STAGE_WITH_USER.bpmn"})
+public class DCU_MIN {
+
+    public static final String DISPATCH = "CallActivity_1rowgu5";
+    public static final String PO_SIGN_OFF = "CallActivity_0pmeblj";
+    public static final String INITIAL_DRAFT = "CallActivity_1ket68y";
```
MarkWrightBjss, line 42: This is what Will described as "smoke and mirrors" in his comment here:
https://github.com/UKHomeOffice/hocs-workflow/pull/348
I've just been renaming the workflow boxes to be more meaningful.
Do we have a better solution to Will's point?

src/test/java/uk/gov/digital/ho/hocs/workflow/processes/DCU_MIN_MinisterSignOff.java
```diff
@@ -0,0 +1,115 @@
...
+
+    }
+
+    @Test
+    public void rejectSignOffScenario() {
+
+        when(dcuMinSignOffProcess.waitsAtUserTask(APPROVE_MINISTER_SIGN_OFF))
+                .thenReturn(task -> task.complete(withVariables("valid", true, "MinisterSignOffDecision", "REJECT")));
+
+        when(dcuMinSignOffProcess.waitsAtUserTask(DCU_MIN_MinisterSignOff.VALIDATE_REJECTION_NOTE))
```
MarkWrightBjss, line 72: The "DCU_MIN_MinisterSignOff" qualifier here feels redundant and inconsistent.

src/test/java/uk/gov/digital/ho/hocs/workflow/processes/DCU_MIN_MinisterSignOff.java
```diff
@@ -0,0 +1,115 @@
...
+
+        verify(dcuMinSignOffProcess)
+                .hasFinished(VALIDATE_REJECTION_NOTE);
+
+        verify(dcuMinSignOffProcess)
+                .hasFinished(DCU_MIN_MINISTER_SIGN_OFF_END);
+
+        verify(dcuMinSignOffProcess, never()).hasFinished(VALIDATE_NOT_APPLICABLE);
+
+        verify(bpmnService).updateAllocationNote(any(), any(), any(), eq("REJECT"));
```
MarkWrightBjss, line 88: Would be good for the third "any()" here to match against the actual case note text.

MarkWrightBjss (changes requested):
I think workflow unit tests come in useful when resolving conflicts in BPMN files. I might have missed it, is there a covering unit test here for the back link on the new screen?

src/test/java/uk/gov/digital/ho/hocs/workflow/processes/DCU_MIN.java
philbarton: Will's point was that we're testing the workflows directly. We can't get away from that unless we change the design. My opinion is 'we are where we are' and any tests are better than non. I wasn't sure of the knock on effect of renaming them, and what would happen if there was a duplicate name?

src/test/java/uk/gov/digital/ho/hocs/workflow/processes/DCU_MIN_MinisterSignOff.java
philbarton: I did think that at first, but in test it comes through empty. I'll have a look and figure out how to that.

src/test/java/uk/gov/digital/ho/hocs/workflow/processes/DCU_MIN_MinisterSignOff.java
philbarton: Sorry, I don't understand.

MarkWrightBjss: when(dcuMinSignOffProcess.waitsAtUserTask(VALIDATE_REJECTION_NOTE))

philbarton: I see...now ta

src/test/java/uk/gov/digital/ho/hocs/workflow/processes/DCU_MIN.java
MarkWrightBjss: We are where we are.
(The modeller prevents entering a duplicate.)

src/test/java/uk/gov/digital/ho/hocs/workflow/processes/DCU_MIN_MinisterSignOff.java
MarkWrightBjss: "VALIDATE_REJECTION_NOTE" should mock a return value for "CaseNote_MinisterReject"; that value should come through in place of the empty. (Same thing for next unit test below.)

src/test/java/uk/gov/digital/ho/hocs/workflow/processes/DCU_MIN_MinisterSignOff.java
philbarton: removed qualifire

src/test/java/uk/gov/digital/ho/hocs/workflow/processes/DCU_MIN_MinisterSignOff.java
philbarton: added rejection string

philbarton:
Also added a back link test.

MarkWrightBjss (approved):
Cheers, good work.

PR #360, opened 2021-01-14
HOCS-2015 : When case converted then remove allocation if user not in new team

MarkWrightBjss:
Create stage validates the provided user is in the provided team.


philbarton (commented):
How is caseworkClient.createStage affected with a null userUUID?

MarkWrightBjss:
Calling createStage with a userUUID is used by "STAGE_WITH_USER" flow.
The "STAGE" flow calls createStage at line 39, which at line 40 calls the above method with a null userUUID.
So setting the userUUID to null at line 52 simulates the request to change stage without a userUUID, resulting in no user being allocated when the case changes stage.

philbarton (approved):
Looks good.

PR #363, opened 2021-01-18
Fix name of project incorrect - was worflow and not workflow

LiamMacP:
Project name was incorrectly spelt worflow instead of workflow in drone.yml `install-docker-image-latest`step. 
This change ensures that the pipeline can be ran successfully.


Paulcr-ho (approved):

zuzak:
(squash this)

LiamMacP:
Was squash merged 👍🏼

PR #364, opened 2021-01-20
HOCS-2358: Revert 2266 to allow for Prod deployment

LiamMacP:
As a result of [PR 355](https://github.com/UKHomeOffice/hocs-workflow/pull/355) effecting old cases and potentially putting them in a state of disrepair, this PR reverts the change until a suitable workaround has been identified.


LiamMacP:
No longer required - different solution was used.

PR #365, opened 2021-01-21
HOCS-2189 : add ability to update cases within transfer stage

abarker66:
Added the capability to alter the case due date, record notes and accept or reject a transferred case.


philbarton:
Hi Andy, We have the ability to unit test workflows now.

abarker66:
Phil, I will merge this into the epic as it is very late - but I agree with you moving forwards we should do this for all camunda work. You are aware that some stages like Triage are actually potentially reentrant - I don't know how the tests would handle this.

philbarton:
> Phil, I will merge this into the epic as it is very late - but I agree with you moving forwards we should do this for all camunda work. You are aware that some stages like Triage are actually potentially reentrant - I don't know how the tests would handle this.

Thats fine, some of the existing tests go round in loops. A mock can generate different output for each loop through.

PR #368, opened 2021-01-25
HOCS-2283 - Fix markup teams being updated when team disabled

ARobinson26:
Fix for when an override markup team has been disabled and the case is then updated with no override team selected. The case should no longer have an override team selected and the _correct_ initial team(s) should be selected.

- Branched from #377 as changes to the `updateTeamSelection` meant I could implement a simpler solution.
- Added a check into `updateTeamSelection` to see if Override Teams are inactive. If they are, remove them by setting null.
- Add tests to `BpmnServiceTest`.

### DIFF
--- a/src/main/java/uk/gov/digital/ho/hocs/workflow/BpmnService.java
+++ b/src/main/java/uk/gov/digital/ho/hocs/workflow/BpmnService.java
@@ -179,17 +179,35 @@ public class BpmnService {
         Map<String, String> teamsForTopic = new HashMap<>();
 
         if (StringUtils.hasText(draftingUUIDString)) {
-            log.info("Overwriting draft team with {}", draftingUUIDString);
             TeamDto draftingTeam = infoClient.getTeam(UUID.fromString(draftingUUIDString));
-            teamsForTopic.put("OverrideDraftingTeamUUID", draftingTeam.getUuid().toString());
-            teamsForTopic.put("OverrideDraftingTeamName", draftingTeam.getDisplayName());
+
+            if (draftingTeam.isActive()){
+                teamsForTopic.put("OverrideDraftingTeamUUID", draftingTeam.getUuid().toString());
+                teamsForTopic.put("OverrideDraftingTeamName", draftingTeam.getDisplayName());
+                log.info("Overwriting draft team with {}", draftingUUIDString);
+            }
+            else {
+                teamsForTopic.put("OverrideDraftingTeamUUID", null);
+                teamsForTopic.put("OverrideDraftingTeamName", null);
+                log.info("Removing Override Drafting Team as selected team was inactive.");
+            }
+
+
         }
 
         if (StringUtils.hasText(privateOfficeUUIDString)) {
-            log.info("Overwriting po team with {}", privateOfficeUUIDString);
             TeamDto pOTeam = infoClient.getTeam(UUID.fromString(privateOfficeUUIDString));
-            teamsForTopic.put("OverridePOTeamUUID", pOTeam.getUuid().toString());
-            teamsForTopic.put("OverridePOTeamName", pOTeam.getDisplayName());
+            if (pOTeam.isActive()){
+                teamsForTopic.put("OverridePOTeamUUID", pOTeam.getUuid().toString());
+                teamsForTopic.put("OverridePOTeamName", pOTeam.getDisplayName());
+                log.info("Overwriting po team with {}", privateOfficeUUIDString);
+            }
+            else{
+                teamsForTopic.put("OverridePOTeamUUID", null);
+                teamsForTopic.put("OverridePOTeamName", null);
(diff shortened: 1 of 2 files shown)


MarkWrightBjss:
"I've raised this as a bug: HOCS-2236" - I'm not sure that HOCS-2236 is the correct reference ticket number. Please confirm.

ARobinsonHOD:
@MarkWrightBjss Apologies, that should have been HOCS-2366 - I've corrected this in the original message.

MarkWrightBjss (changes requested):
Would it be possible to add some workflow unit tests to cover the changes?

src/test/java/uk/gov/digital/ho/hocs/workflow/processes/DCU_DTEN_MARKUP.java
```diff
@@ -0,0 +1,123 @@
...
+    public static final String DRAFTING_UKVI_BF_IE_Ministerial = "d0f4d3a4-357f-421b-afb1-cdfd3970c674";
+    public static final String HOME_SEC_OFFICE = "3d2c7893-92c5-4347-804a-8826f06f0c9d";
+    public static final String MINISTER_FOR_LORDS = "5311138e-33bc-434d-9bce-a933a51a3146";
+    public static final String CASE_UUID = "ABC";
+
+
+    @Rule
+    @ClassRule
+    public static TestCoverageProcessEngineRule rule = TestCoverageProcessEngineRuleBuilder.create()
+            .assertClassCoverageAtLeast(0.86)
```
MarkWrightBjss, line 54: Change this to 0.2, i.e.
.assertClassCoverageAtLeast(0.2)

src/test/java/uk/gov/digital/ho/hocs/workflow/processes/DCU_MIN_MARKUP.java
```diff
@@ -0,0 +1,137 @@
...
+    public static final String DRAFTING_UKVI_BF_IE_Ministerial = "d0f4d3a4-357f-421b-afb1-cdfd3970c674";
+    public static final String HOME_SEC_OFFICE = "3d2c7893-92c5-4347-804a-8826f06f0c9d";
+    public static final String MINISTER_FOR_LORDS = "5311138e-33bc-434d-9bce-a933a51a3146";
+    public static final String CASE_UUID = "ABC";
+
+
+    @Rule
+    @ClassRule
+    public static TestCoverageProcessEngineRule rule = TestCoverageProcessEngineRuleBuilder.create()
+            .assertClassCoverageAtLeast(0.86)
```
MarkWrightBjss, line 68: Change this to 0.2, i.e.
.assertClassCoverageAtLeast(0.2)

src/test/java/uk/gov/digital/ho/hocs/workflow/processes/DCU_TRO_MARKUP.java
```diff
@@ -0,0 +1,110 @@
...
+    public static final String DRAFTING_UKVI_BF_IE_Ministerial = "d0f4d3a4-357f-421b-afb1-cdfd3970c674";
+    public static final String HOME_SEC_OFFICE = "3d2c7893-92c5-4347-804a-8826f06f0c9d";
+    public static final String MINISTER_FOR_LORDS = "5311138e-33bc-434d-9bce-a933a51a3146";
+    public static final String CASE_UUID = "ABC";
+
+
+    @Rule
+    @ClassRule
+    public static TestCoverageProcessEngineRule rule = TestCoverageProcessEngineRuleBuilder.create()
+            .assertClassCoverageAtLeast(0.86)
```
MarkWrightBjss, line 54: Change this to 0.2, i.e.
.assertClassCoverageAtLeast(0.2)

MarkWrightBjss (changes requested):
These changes should get the tests passing. (I think each PR should be passing regardless of other branches.)

MarkWrightBjss (changes requested):
The TRO flow looks good. For DTEN and MIN case types, the PR looks good, but I think the problem will still exist for FAQs.

Adding additional backend calls from Camunda to Java in the FAQ flow would be an option to fix this, but there's already quite a few additional calls being added here.

Instead of having all these additional workflow tasks hitting a new `updateTeamIfInactive` endpoint, would it be possible to modify the existing `updateTeamSelection` to do the new `active` check, with the new processing as needed? It looks like `updateTeamSelection` is always called after the new endpoint anyway, plus it'll be fixed anywhere else that the team selection is updated, current or future.

ARobinsonHOD:
As suggested by @MarkWrightBjss, I've moved the logic into `updateTeamSelection` rather than a separate workflow step and function. I've rebased from #377 as the changes in this PR simplified the changes needed considerably.

agraham19 (approved):

MarkWrightBjss (approved):

PR #369, opened 2021-01-25
HOCS-1907 : Change MPAM workflow to remove the unsed Confirm and Save changes radio butons from the business area change screen.

philbarton:
There are 6 locations for the screen 3 x triage and 3 x draft. In each case the output of the call activity nas been defaulted to "Confirm".

### DIFF
--- a/src/main/resources/application-auto-deployment.properties
+++ b/src/main/resources/application-auto-deployment.properties
@@ -1,0 +1,1 @@
+camunda.bpm.auto-deployment-enabled=true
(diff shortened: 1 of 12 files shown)


src/main/resources/application-auto-deployment.properties
```diff
@@ -0,0 +1 @@
+camunda.bpm.auto-deployment-enabled=true
```
LiamMacP, line 1: Is this doing anything? It's default is true but the option is disabled when using 'SpringBootProcessApplication'. If this is needed could it not be in the application.properties as specify other camunda props there.
https://docs.camunda.org/manual/7.14/user-guide/spring-boot-integration/configuration/#camunda-engine-properties

philbarton: That documentation is confusing. From what I've seen It defaults to true unless when using 'SpringBootProcessApplication' were it defaults to false.
By setting this to true, a restart of the service re-loads the workflows into the DB, saves dropping the schema.
Thats the behaviour I see if I add the profile 'auto-deployment' to my local startup.
Spring looks for property files named after the profile, so it gets picked up.

LiamMacP (approved):
Changes look good to me...

PR #370, opened 2021-01-25
HOCS-2296 : change to Official clears Ministerial data inc Escalated, fix

MarkWrightBjss:
Changing a case from reftype "Ministerial" to "Official" clears ministerial only data values: MinSignOffTeam, Addressee
At both Triage Escalated and Draft Escalated.


src/main/resources/processes/MPAM_TRIAGE_ESCALATE.bpmn
```diff
@@ -368,16 +373,16 @@
         <di:waypoint x="1613" y="710" />
       </bpmndi:BPMNEdge>
       <bpmndi:BPMNEdge id="SequenceFlow_19l3zia_di" bpmnElement="SequenceFlow_19l3zia">
-        <di:waypoint x="1068" y="2170" />
-        <di:waypoint x="1068" y="2255" />
-        <di:waypoint x="278" y="2255" />
+        <di:waypoint x="1068" y="1831" />
```
zuzak, line 376: All this change, and many of the next few touched lines, does is move the box slighty. I'm wondering if we should start getting into the practice of only applying chunks (`git add -p`) that have meaningful changes, instead of blindly adding the entire file to a commit?

LiamMacP: That might be a little difficult to work out what waypoints only change the one piece people have touched. Might be all these are needed. I'd also argue that for workflow changes we should only be approving after pulling the branch and actually looking at it

LiamMacP (approved):
Changes look good - tests are good

PR #372, opened 2021-01-26
HOCS-2375 - Wrap call activity mocks in a simpler interface.

philbarton:
Test utility to simplify callActivity mocking.

Refactored code to use new utility class.


SeanReedHomeOffice (approved):
Looks good.

PR #373, opened 2021-01-29
HOCS-2027: MPAM global permissions

parth-ho:


philbarton (changes requested):
The tests need to be corrected.

philbarton (approved):
Looks good. There is a lot here, so thorough integration testing will be required.

PR #374, opened 2021-02-01
HOCS-2398 : update bpmn diagrams to allow escalated stages to move ca…

abarker66:
This update to the top level, triage escalation and draft escalation diagrams allows a user to send a case back to the Transfer stage from the Triage Escalated stage and the Draft Escalated stage.

### DIFF
--- a/src/main/resources/processes/MPAM.bpmn
+++ b/src/main/resources/processes/MPAM.bpmn
@@ -684,6 +684,7 @@
         <camunda:out source="RefTypeStatus" target="RefTypeStatus" />
         <camunda:out source="RefType" target="RefType" />
         <camunda:out source="LastUpdatedByUserUUID" target="CampaignUserUUID" />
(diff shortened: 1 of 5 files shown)


LiamMacP (commented):
Looks to be a good start, couple of things though: 

- This will need to go in after [HOCS-2406](https://github.com/UKHomeOffice/hocs-workflow/pull/376#pullrequestreview-581667384) as it will need to clear out the assigned user. 
- Might need to look at clearing the business unit on this change (although I will look at if there is a bigger issue with this)

LiamMacP:
There are another set of {Triage,Draft} Escalated stages within MPAM - does this need applying to them?

abarker66:
No because you can't change the BusArea in them - they are fixed

LiamMacP (approved):
Pulled branch and visually checked that code goes through the clear user. Tests look to cover this change. Approved.

PR #376, opened 2021-02-01
HOCS-2406 unallocate user when moving back to transfer stage

abarker66:
Added an activity to clear the userid (unallocate) when transferring cases back to transfer stage from triage and draft stages.

### DIFF
--- a/src/main/resources/processes/MPAM.bpmn
+++ b/src/main/resources/processes/MPAM.bpmn
@@ -1005,8 +1005,7 @@
       </bpmn:extensionElements>
       <bpmn:incoming>Flow_1tjgt44</bpmn:incoming>
       <bpmn:incoming>Flow_0fz9w03</bpmn:incoming>
-      <bpmn:incoming>Flow_1gvq3nw</bpmn:incoming>
-      <bpmn:incoming>Flow_15cziqn</bpmn:incoming>
+      <bpmn:incoming>Flow_0ded6a7</bpmn:incoming>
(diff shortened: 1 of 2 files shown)


LiamMacP:
Worth adding some tests for this? This will also effect old cases, although not necessarily a blocker 🙂

abarker66:
All the current bpmn processes tests pass. Don't have the time given the bug backlog.

philbarton:
> All the current bpmn processes tests pass. Don't have the time given the bug backlog.

Our definition of done states we should have unit tests.

Without tests the next set of manual changes will likely break these changes and cause more regression bugs.

abarker66:
'Our definition of done states we should have unit tests.' - since we have no detailed software requirements against which to test all we are doing is testing the code does what the code does and altering the tests to accommodate the functional behaviour of the code as it is. Pointless.  We should and could have a set of end to end functional tests to replace all the bpm service tests which are a set of laughable mockito vanity tests which prove nothing.

philbarton:
> 'Our definition of done states we should have unit tests.' - since we have no detailed software requirements against which to test all we are doing is testing the code does what the code does and altering the tests to accommodate the functional behaviour of the code as it is. Pointless. We should and could have a set of end to end functional tests to replace all the bpm service tests which are a set of laughable mockito vanity tests which prove nothing.

They find merge issues in a manual merge process.

LiamMacP (approved):
Changes look good - tests added to aid future merge verification.

philbarton:
This was merged without taking into account my comments. These tests aren't accurate. They miss half of the flows.

PR #378, opened 2021-02-09
HOCS-2463 replace activities removed in merge

abarker66:
This work replaces activities removed in a recent merge activity. Also added some tests to ensure this does not happen again.


philbarton (approved):
Looks good.
**assertClassCoverageAtLeast** could be set to 0.67 because coverage has increased.

PR #379, opened 2021-02-16
Hocs 2471 - draft/po team override

philbarton:
Changes to take into account any overridden drafting or po team, when associating with work stacks. 
Added a new test approach for Call activity variable checking.


MarkWrightBjss (approved):

philbarton:
Workflow changes - consider override teams when adding to work stack

PR #380, opened 2021-02-17
HOCS-2225 : DCU PO change topic, fix

MarkWrightBjss:
After getting the new team for the topic, the case is now allocated to the new PO Team, and then routed through the deadline re-set flow, because HO cases have a different SLA.


SeanReedHomeOffice (approved):
LGTM

PR #382, opened 2021-02-23
Hocs 2471 draft team override

philbarton:
Defect corrected by changing from null to "" 

Manually tested for both override and for no-override, for both draft ream and po team.


abarker66 (approved):
All looks good. The escape sequence looks like a backslash in quotes until I looked closer, all good.

PR #386, opened 2021-03-08
HOCS:2283 Set override values to empty rather than null

ARobinson26:
In a previous change, I set override team values to `null` if they were a disabled team. This caused issues setting the active team for the case, as there's a check in `DCU_MIN.bpmn` that sets the active team to the default topic team if `overrideTeamUuid = ""`.
This change sets the override teams to empty strings rather than null values.


philbarton (approved):
Looks good.

PR #388, opened 2021-03-17
HOCS-2539 : add top level bpmn digram for FOI

abarker66:
This ticket adds the initial top level bpmn digram for FOI workflow. It does nothing but close a created case. Further stages shall be added to this diagram.
A process test is also included.

### DIFF
--- a/src/test/java/uk/gov/digital/ho/hocs/workflow/processes/FOI.java
+++ b/src/test/java/uk/gov/digital/ho/hocs/workflow/processes/FOI.java
@@ -1,0 +1,61 @@
+package uk.gov.digital.ho.hocs.workflow.processes;
+
+import org.camunda.bpm.engine.test.Deployment;
+import org.camunda.bpm.engine.test.ProcessEngineRule;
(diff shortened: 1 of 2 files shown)


src/test/java/uk/gov/digital/ho/hocs/workflow/processes/FOI.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Test
+    public void happyPath() {
+
+        Scenario.run(FOIProcess)
+                .startByKey("FOI")
+                .execute();
+
+        verify(FOIProcess)
+                .hasCompleted("Activity_1d2ue6g");
```
SeanReedHomeOffice, line 55: Can we use a meaningful name instead of Activity_1d2ue6g ?

abarker66: That's an internal id used by camunda, it is given a proper real name but the tests use the internal id.

SeanReedHomeOffice (approved):
LGTM

PR #394, opened 2021-04-06
HOCS-2695 : CCT workflow

MarkWrightBjss:
Stages for CCT workflow skeleton.
Merge to epic.


MarkWrightBjss:
Each flow is for a completely different stage. At the moment they are very similar as they're starting points for each stage. I expect some divergence and growth as user requirements come through.

robertdeniszczyc2 (approved):

PR #399, opened 2021-04-21
HOCS-2769 Deploy to hocs-delta automatically

zuzak:
This adds a step to Drone that fires on push to the COMP epic branch,
which triggers a deployment to the hocs-delta environment.

Previously we were deploying manually to hocs-delta, and using a fixed
branch-based tag, which was causing inconsistencies and confusion. Here
we deploy a tag based on the Drone build number, which is easier to
track. If we wanted, we could deploy using the SHA hash for a similar
effect.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -142,6 +142,25 @@ steps:
     depends_on:
       - clone kube repo
 
+  - name: deploy to delta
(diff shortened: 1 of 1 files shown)


philbarton (approved):
Looks good to me.

PR #400, opened 2021-04-22
HOCS-2739 : CCT "is template" cases skip Triage

MarkWrightBjss:
New decision point to check if "IsTemplate" checkbox has been ticked ("TL"); if so, flows to CCT Response (Draft) stage, else continues to Triage.
Merge to epic.


robertdeniszczyc2 (approved):
Looks good to me

PR #403, opened 2021-04-26
HOCS-2753 User selects No at the acceptance stage

agraham19:
* Update Camunda flows
* Update tests


src/main/resources/processes/FOI_ACCEPTANCE.bpmn
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
```
SeanReedHomeOffice, line 1: I think Camunda and casework should be updated in a single call the BpmnService. That seems to be the existing pattern.

SeanReedHomeOffice, line 1: On the out mapping here, can't we just have AcceptanceTeam instead of ${execution.getVariable("AcceptanceTeam")} ?

SeanReedHomeOffice: hmm, I am having second thoughts. Not sure the best pattern here.

agraham19: Would you suggest creating a new method in the BPMNService to handle this?

agraham19: I've made this change and it still works as expected.

SeanReedHomeOffice: That's what I was wondering. Also shouldn't we also be wiping the 'Directorate' Camunda process variable as we clear this in the casework db?

agraham19: I've created a new method to do this in one task and clear both variables.

SeanReedHomeOffice (approved):
LGTM

PR #404, opened 2021-04-26
HOCS-2819 : Acceptance deadline

SeanReedHomeOffice:
A Camunda timer is used for the 48 hour deadline. Specifically an interrupting timer boundary event is used. See  https://docs.camunda.io/docs/reference/bpmn-workflows/timer-events/timer-events/
The deadline is handled entirely within the Acceptance stage.
After the deadline passes the user is presented with a new screen explaining the deadline has expired.
The deadline is calculated using the Info micro-service.


abarker66:
Is it necessary to inject a Clock object? the LocalDate class has built in static methods to give you now() time already.

SeanReedHomeOffice:
> Is it necessary to inject a Clock object? the LocalDate class has built in static methods to give you now() time already.

The code is impossible to unit test without introducing the Clock. Using the clock allows you to fix what time is 'now'.

abarker66:
Ok, I thought it was to allow testing - fair enough.

agraham19:
You could change the Camunda expression "${bpmnService.calculateDeadline('FOI', 2)}" to pass in a "now" datetime object and get the date from that. That way when you test it, you could pass a fixed date instead.

agraham19 (approved):

PR #405, opened 2021-04-28
HOCS-2885 Automatically deploy to gamma on push

zuzak:
This adds a step to Drone that fires on push to the FOI epic branch,
which triggers a deployment to the hocs-gamma environment.

Previously we were deploying manually to hocs-gamma, and using a fixed
branch-based tag, which was causing inconsistencies and confusion. Here
we deploy a tag based on the Drone build number, which is easier to
track. If we wanted, we could deploy using the SHA hash for a similar
effect.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -104,6 +104,25 @@ steps:
         - push
         - promote
 
+  - name: deploy to gamma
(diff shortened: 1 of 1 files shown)


parth-ho (approved):
LGTM

PR #406, opened 2021-04-29
HOCS-2670 Draft team Acceptance (MVP)

agraham19:
* Add new stage draft team acceptance stage

### DIFF
--- a/src/main/resources/processes/FOI_ACCEPTANCE.bpmn
+++ b/src/main/resources/processes/FOI_ACCEPTANCE.bpmn
@@ -9,14 +9,14 @@
       <bpmn:outgoing>Flow_0eg5k1f</bpmn:outgoing>
     </bpmn:userTask>
     <bpmn:endEvent id="END_FOI_ACCEPTANCE" name="End Stage">
-      <bpmn:incoming>Flow_0ax0fmd</bpmn:incoming>
       <bpmn:incoming>Flow_1unjlmz</bpmn:incoming>
+      <bpmn:incoming>Flow_1es0j8i</bpmn:incoming>
     </bpmn:endEvent>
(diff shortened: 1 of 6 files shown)


src/main/resources/processes/FOI_DRAFT_ACCEPTANCE.bpmn
```diff
@@ -0,0 +1,107 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
SeanReedHomeOffice, line 1: This stage should just be called FOI_ DRAFT or FOI_CONSIDER_AND_DRAFT

src/main/resources/processes/FOI_ACCEPTANCE.bpmn
```diff
@@ -9,14 +9,15 @@
       <bpmn:outgoing>Flow_0eg5k1f</bpmn:outgoing>
```
SeanReedHomeOffice, line 9: About ${execution.getVariable('DraftAcceptCase') != null && DraftAcceptCase == "N"}.
Not sure that DraftAcceptCase is the right variable to use - it is the output of a user task in a future stage. What we need to check is has this case already been accepted by the acceptance team. So I think maybe we should set a variable (date accepted?) in this stage when a case is accepted, and test for the presence of that variable.

SeanReedHomeOffice (commented):
Looks good, a couple of comments

src/main/resources/processes/FOI_ACCEPTANCE.bpmn
agraham19: I have made this change

src/main/resources/processes/FOI_DRAFT_ACCEPTANCE.bpmn
agraham19: I have made this change

parth-ho (approved):

PR #408, opened 2021-05-04
HOCS-2739: moving WCS team update case

LiamMacP:
As we are currently changing the internal representation of the data, 
we need to pull this out of the stage when we finish one. As we 
change it for the stage level we can assign the outer variable to the 
same.

Currently when a team is moved outside a workflow, the internal 
representation of the case is not update-able. This change adds an 
endpoint that allows for the updating of the data. This endpoint is 
protected by the `@Authorised` annotation for users with write 
permissions.


MarkWrightBjss (changes requested):
Because the majority of WCS cases have now been created, and are just getting passed around different case working teams as load allows, changes to WCS.bpmn will not take affect.
Unfortunately, looking at the diff, I can't tell if the changes are required, or if they're because you opened the file in the latest version of the model, or something else.

MarkWrightBjss (approved):
WCS have been told this will only work for new cases.

PR #410, opened 2021-05-05
HOCS-2923: allow enquiry reason subject change

LiamMacP:
This change encompasses the following changes: 

- Now load all bpmn files in the processes folder and any subfolder. Allowing for seperation of workflows under business unit.
- Pull out the Update Enquiry Subject and Reason from MPAM Triage and put into it's own workflow. This is then reused in the MPAM Triage Esclate.


SeanReedHomeOffice (approved):
LGTM. Well done on introducing this approach.

philbarton (approved):
Looks good to me.

PR #413, opened 2021-05-06
HOCS-2914 back button workflow handling

LiamMacP:
Allow for the back button action on the 
`UserTask_RequestContributionInput` screen incase users 
mistakenly click to go down that path.


philbarton (approved):
Looks good

PR #416, opened 2021-05-07
Feature/hocs 2944 cch returned cases ws

MarkWrightBjss:
New "Reopened" stage.
Cases re-opened from "Closed" stages are now sent here, instead of "Returned" ws.
Merge to epic.


philbarton (approved):
Looks good, tested locally

PR #425, opened 2021-05-17
HOCS-2778: Comp Creation tests

ARobinson26:
Add tests for `COMP_CREATION.bpmn`


MarkWrightBjss:
Should the screen contents be covered by HOCS-2778 ?

philbarton (approved):
Your coverage is actually 0.88888888 :) 
What I do is set it to 99 and then let it fail, but use the new number as the threshold.
Not sure if you were aware but the coverage is also visualised here .....  open target/process-test-coverage/uk.gov.digital.ho.hocs.workflow.processes.COMP_CREATION/COMP_CREATION.html

philbarton:
> Should the screen contents be covered by HOCS-2778 ?

Good point. Not sure.

ARobinson26:
Thanks @philbarton - I'll take a look at the visual, didn't know about that!

Going to look at screen content tests as suggested by @MarkWrightBjss, now.

PR #427, opened 2021-05-18
HOCS-3029 : CCT Response team options: escalate, and return

MarkWrightBjss:
New stage for "Response Escalated".
New route from CCT Response back to CCT Triage.
Team allocations for these new routes.
Merge to epic.


robertdeniszczyc2 (approved):
Looks ok to me

PR #430, opened 2021-05-21
Feature/hocs 3044 comp type fix

MarkWrightBjss:
Give each stage a uniquely named "CompType" value, so they don't interfere with each other.
Merge to epic.


philbarton (approved):
Looks good to me, might need a change if a comment on the other PR is applied.

PR #436, opened 2021-05-26
HOCS-3027 : CCT Stage 2 complete case

MarkWrightBjss:
"Out" params on both Stages that change "Stage" variable.
Check for "Stage2" after CCT flow; direct to completion for "Stage2".
Merge to epic.


philbarton (approved):
Looks good to me.

PR #446, opened 2021-06-04
HOCS-2775 : COMP unit tests

MarkWrightBjss:
Unit test coverage for top level COMP.bpmn workflow.
I've renamed some tasks away from the auto-generated random strings.
Also includes just a single unit test for Minor, which hasn't been agreed yet.
Merge to epic.


gardnersam (approved):
I like the renaming, the random strings were confusing

PR #448, opened 2021-06-09
HOCS-3098 FOI Auto allocation of case at Data Input

agraham19:
* Automatically allocate FOI cases to their creator when the data input stage is entered


SeanReedHomeOffice (approved):
LGTM

PR #453, opened 2021-06-15
HOCS-2545 FOI Dispatch (MVP)

agraham19:
* Add Dispatch stage to FOI
* Add Soft Close Stage to FOI


abarker66 (approved):
Looks good, we now have new dateOfResponse

PR #454, opened 2021-06-16
HOCS-3193 Sticky Cases

SeanReedHomeOffice:
If at the end of a stage the user is a member of the team that handles the case at the next stage, then the case might want to be be allocated to them automatically and displayed on screen, instead of just going to the workstack.

When the workflow service - updateStageForward() - detects that the end of a stage is reached - in getStage() - it checks if the current user is on the team for the next stage. If they are the user is allocated to that case, and data relating to the new stage is returned to the front end.
The front end code detects if the stageId has changed, and uses that instead when it redirects.

The feature can be turned on by setting a process variable in the Camunda diagram. 

sticky_cases001


agraham19 (approved):

gary-newbolt:
lgtm too thanks @SeanReedHomeOffice

gary-newbolt:
@SeanReedHomeOffice lgtm thanks

LiamMacP (approved):
Changes look good to me, you will need to rebase this branch back off main to bring it upto date and keep linear history. Also probably worthwhile changing the PR title for when you merge it in 👍🏽

PR #455, opened 2021-06-17
HOCS-3189 Building with gradle 7.1 and test dependencies up to date

gardnersam:


build.gradle
```diff
@@ -1,7 +1,7 @@
 buildscript {
 	ext {
 		springBootVersion = '2.1.3.RELEASE'
```
LiamMacP, line 3: This probably should be brought upto date aswell. Although that could be more of a pain due to having to follow the migration guides.

gardnersam: I agree, it's worth at least seeing how much of a pain it is. If we can move up several versions pain free at least we may aswel even if it means pausing before the most up to date version. This PR was mostly to check that it still built outside of my local env before pushing on

philbarton (approved):
Looks good to me

PR #460, opened 2021-07-02
HOCS-3087 Minimised docker image size and added custom healthcheck

gardnersam:


Dockerfile
```diff
@@ -1,19 +1,17 @@
-FROM quay.io/ukhomeofficedigital/openjdk11:v11.0.5_10
+FROM quay.io/ukhomeofficedigital/alpine:3.10
 
 ENV USER user_hocs_workflow
 ENV USER_ID 1000
```
zuzak, line 4: Do you need these declarations now?

Dockerfile
```diff
@@ -1,19 +1,17 @@
...
 ENV GROUP group_hocs_workflow
 ENV NAME hocs-workflow
 ENV JAR_PATH build/libs
 
-RUN yum update -y glibc && \
-    yum update -y nss && \
-    yum update -y bind-license
+RUN apk add openjdk11-jre
```
philbarton, line 9: Can this be a specific version v11.0.5_10 ?

philbarton (changes requested):
Looks good to me, just a comment on version

Dockerfile
```diff
@@ -1,19 +1,17 @@
-FROM quay.io/ukhomeofficedigital/openjdk11:v11.0.5_10
+FROM quay.io/ukhomeofficedigital/alpine:3.10
```
philbarton, line 1: I wonder wether we should have a new base image that has the health check in it?

philbarton (changes requested):

Dockerfile
gardnersam: It is specifically 11.0.4_p4-r1 despite the unclear package name, see https://pkgs.alpinelinux.org/package/v3.10/community/s390x/openjdk11-jdk

philbarton: is there a later alpine version that matchs  v11.0.5_10? 
Just thinking about maintaining versions.

Dockerfile
gardnersam: USER_ID isn't needed, removed

Dockerfile
```diff
@@ -1,19 +1,18 @@
...
 ENV USER user_hocs_workflow
-ENV USER_ID 1000
 ENV GROUP group_hocs_workflow
 ENV NAME hocs-workflow
 ENV JAR_PATH build/libs
 
-RUN yum update -y glibc && \
-    yum update -y nss && \
-    yum update -y bind-license
+USER root
```
gardnersam, line 8: We have to swap to root here as the image is set up to make us a user without permission to install, see https://quay.io/repository/ukhomeofficedigital/alpine/manifest/sha256:81f32a5fca1f3f053703c0c2488fbd6b6bed4f5af8108a2ae186095d34f8c9e6 for more info on the alpine image

Dockerfile
gardnersam: Moved to alpine 3.13. which gives us Java 11.0.9_11. Nothing in the release notes stand out to me as an obvious problem

Dockerfile
gardnersam: My personal preference would be to have the clean base and build on it like this but happy to be overruled, I think images with whatever added are hard to interpret for future people

philbarton: lets leave it clean. Shame we have to add the healthcheck to 9 repos!!!

Dockerfile
philbarton: Sounds good.

scripts/HealthCheck.java
```diff
@@ -0,0 +1,25 @@
+package uk.gov.digital.ho.hocs;
+
+import java.io.IOException;
+import java.net.URI;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+
+public class HealthCheck {
```
zuzak, line 9: Can you add a comment here explaining what `HealthCheck` is for? (i.e. make it clear this is just for docker-compose and not for k8s health checks)

scripts/HealthCheck.java
```diff
@@ -0,0 +1,25 @@
...
+        var client = HttpClient.newHttpClient();
+        var request = HttpRequest.newBuilder()
+                .uri(URI.create("http://localhost:8080/actuator/health"))
+                .header("accept", "application/json")
+                .build();
+
+        HttpResponse<String> response = client.send(request, HttpResponse.BodyHandlers.ofString());
+
+        if (response.statusCode() != 200 || !response.body().contains("UP")) {
+            throw new RuntimeException("Healthcheck failed");
```
zuzak, line 21: Would be good if this message explained why it failed. Or is the stack trace sufficient?

scripts/HealthCheck.java
```diff
@@ -0,0 +1,25 @@
...
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
+
+public class HealthCheck {
+
+    public static void main(String[] args) throws IOException, InterruptedException {
+        var client = HttpClient.newHttpClient();
+        var request = HttpRequest.newBuilder()
+                .uri(URI.create("http://localhost:8080/actuator/health"))
```
zuzak, line 14: Will this always be 8080?

gardnersam: 8080 is the spring boot default so unless someone goes and changes it on purpose yes. This doesn't tie us to a port any more than having it defined as 8080 in the curl based health check

scripts/HealthCheck.java
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Class used for container health check by docker compose
+ */
+public class HealthCheck {
+
+    public static void main(String[] args) throws IOException, InterruptedException {
+        var client = HttpClient.newHttpClient();
+        var request = HttpRequest.newBuilder()
+                .uri(URI.create("http://localhost:8080/actuator/health"))
```
LiamMacP, line 17: We pass through a `server.port` through application properties, could that be used here? https://github.com/UKHomeOffice/hocs-workflow/blob/8c15c770c1dc710c58c951495ed26181c7630071/src/main/resources/application.properties#L5

philbarton: Thats the external port. I think we always use 8080 inside the docker-compose context. We also have it hardcode in the previous docker-compose.
Having it as a arg to the HealthCheck would make it more reusable......if it wasn't duplicated anyway!

LiamMacP: I'd argue that we should be defensive with this, we have the property there, there is a chance it could change and it wouldn't be too much to do 🙂

gardnersam: I think the problem with using the application.properties is that 8091 isn't the port we want to health check on, 8080 is.

LiamMacP: The port we want to listen on is the one specified through that `server.port`. We override that in the docker-compose/ kube-deployment to 8080 (but this can be changed to anything).

LiamMacP: I'm happy to be overruled, just wanted to raise it as a potential improvement 🙂

gardnersam: HealthCheck.java now takes the address to check as a command line argument so we can customise in the docker compose file as needed

philbarton: 👍🏻

philbarton (approved):
Looks good to me

PR #463, opened 2021-07-07
Feature/hocs 3326 add server shutdown graceful

philbarton:
- Add server shutdown graceful
- Remove app name from logging


gardnersam (approved):
Looks good

PR #465, opened 2021-07-08
HOCS-3251 save foi topic in data input stage

abarker66:
Added a step to add the primary topic to the current case in the FOI Data Input stage


gary-newbolt:
Hi @abarker66, the workflow phase to register the Primary Topic doesn't seem to activate, nor is it in the case. That being said, the selected `Topics` is saved fine,  it's just the `primaryTopic` that isn't. Furthermore, as it's not part of the new schema defined for FOI Data Input, so there's no visibility of the variable after submission. I think it's worth asking what value this holds for FOI and potentially we could remove the workflow stage entirely.

abarker66:
I rebuilt the workflow and it does save the primary topic and it is in the case. The data input stage will likely change but the FOI Topic defined in this ticket) can be reinserted in the updated stage/screen.

gary-newbolt (approved):

PR #469, opened 2021-07-15
Update case deadline

agraham19:
* Update the case deadline when the received date is update in the creatio…n stage


gary-newbolt (approved):
Great! Cheers @agraham19

PR #472, opened 2021-07-19
HOCS-3399: default path update enquiry reason

LiamMacP:
A live service issue has occurred in the 
`MPAM_UPDATE_ENQUIRY_SUBJECT_REASON` flow whereby a 
value for the direction was provided that was not `FORWARD` or 
`BACKWARD`. This change adds the default flow, so that if the 
direction is not forward it always goes back.


philbarton (approved):
lgtm

LiamMacP:
Closed as no longer required, action taken on another PR.

PR #473, opened 2021-07-19
HOCS-3402 Rejected case at the acceptance stage should allocate case …

agraham19:
…back to FOI allocation team

* Erase allocation team so Workflow defaults to correct team for the allocation stage


gary-newbolt (approved):
lgtm

PR #475, opened 2021-07-20
Feature/hocs 3384 move topics to case create

gary-newbolt:


LaurieBennett (changes requested):
Change variable name and this will pass

src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
```diff
@@ -97,20 +104,44 @@ public void createDocument(UUID caseUUID, List<DocumentSummary> documents) {
...
-            SchemaDto schemaDto = infoClient.getSchema(screenName);
-            List<HocsFormField> fields = schemaDto.getFields().stream().map(HocsFormField::from).collect(toList());
-            List<HocsFormSecondaryAction> secondaryActions = schemaDto.getSecondaryActions().stream().map(HocsFormSecondaryAction::from).collect(toList());
-            fields = HocsFormAccordion.loadFormAccordions(fields);
-            HocsSchema schema = new HocsSchema(schemaDto.getTitle(), schemaDto.getDefaultActionLabel(), fields, secondaryActions, schemaDto.getProps());
-            HocsForm form = new HocsForm(schema, inputResponse.getData());
-            return new GetStageResponse(stageUUID, inputResponse.getReference(), form);
-        } else {
+        if (screenName.equals("FINISH")) {
+            Optional<StageDto> maybeActiveStage = caseworkClient.getActiveStage(caseUUID);
```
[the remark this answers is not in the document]
gary-newbolt: Okay... I wouldn't want to change it. With the type being an `Optional` type, potentially the value could be null. `Optional` is a special Java class that contains a field of the same type as the generic passed in and often used on things such as this where a http call is made with no guarantee that you will get a result. In the subsequent lines, the variable is checked for existence, and if it exists, a more explicit name is given to the extracted object. So I think the term `maybe...` in this instance is appropriate.

src/main/java/uk/gov/digital/ho/hocs/workflow/api/dto/GetStageResponse.java
```diff
@@ -3,12 +3,14 @@
...
 import lombok.AllArgsConstructor;
 import lombok.Getter;
+import lombok.ToString;
 import uk.gov.digital.ho.hocs.workflow.domain.model.forms.HocsForm;
 
 import java.util.UUID;
 
 @AllArgsConstructor
 @Getter
+@ToString
```
gary-newbolt, line 13: This should be taken out though.

LaurieBennett: Agreed

src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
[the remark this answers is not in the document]
LaurieBennett: I figured it was showing some of the merged changes as well, that sounds reasonable, from first look it sounded a bit too much like a temporary variable name, but after further discussion I'm fine with this.

LaurieBennett (approved):
Approved, PASS

src/main/java/uk/gov/digital/ho/hocs/workflow/api/dto/GetStageResponse.java
gary-newbolt: Removed

PR #476, opened 2021-07-22
HOCS-3289 Add Correspondent Details to Case Creation Step

LaurieBennett:
Adding in the ability for a user to enter the correspondent details when creating a case, this includes showing the correct details depending on the correspondence type.

### DIFF
--- a/src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
+++ b/src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
@@ -18,6 +18,8 @@ import uk.gov.digital.ho.hocs.workflow.client.infoclient.dto.TeamDto;
 import uk.gov.digital.ho.hocs.workflow.client.infoclient.dto.UserDto;
 import uk.gov.digital.ho.hocs.workflow.domain.exception.ApplicationExceptions;
 import uk.gov.digital.ho.hocs.workflow.domain.model.forms.*;
+import uk.gov.digital.ho.hocs.workflow.migration.MigrationCaseworkClient;
+import uk.gov.digital.ho.hocs.workflow.migration.MigrationCreateCaseworkCorrespondentRequest;
 import uk.gov.digital.ho.hocs.workflow.util.UuidUtils;
 
 import java.time.LocalDate;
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
```diff
@@ -78,6 +110,13 @@ public CreateCaseResponse createCase(String caseDataType, LocalDate dateReceived
             seedData.putAll(data);
             camundaClient.startCase(caseUUID, caseDataType, seedData);
 
+            //Create correspondent
+            if (correspondentRequest != null) {
+                UUID stageUUID = migrationCaseworkClient.getStageUUID(caseUUID);
+                UUID correspondentUUID = migrationCaseworkClient.saveCorrespondent(caseUUID, stageUUID, correspondentRequest);
```
gary-newbolt, line 116: In the line here... we're using a service that's been written for "Migration" purposes. I think you should switch this out for a dedicated `saveCorrespondence` method. I know you would have to write it, however using this doesn't give any ambiguous use cases. 

So can I request you replace these methods and the use of the migrationCaseworkClient please with methods in the caseworkClient that call the equivalent endpoints in the caseworker service:

`/active-stage/case/{caseUUID}` and `/case/{caseUUID}/stage/{stageUUID}/correspondent`

In the latter endpoint's method, it automatically assigns the primary user so you won't need the next line.

src/test/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowServiceTest.java
```diff
@@ -39,6 +40,9 @@
     @Mock
     private InfoClient infoClient;
 
+    @Mock
```
gary-newbolt, line 43: If you follow the advice above you won't need this.

src/test/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowServiceTest.java
```diff
@@ -50,7 +54,8 @@ public void beforeTest() {
                 caseworkClient,
                 documentClient,
                 infoClient,
-                camundaClient);
+                camundaClient,
+                migrationCaseworkClient);
```
gary-newbolt, line 58: again if you follow the advice above you won't need this either.

gary-newbolt (commented):
I think you should write some tests to cover the scenario where the payload contains correspondence details as there's some logic that is completely untested at it stands.

gary-newbolt (approved):
Great, lgtm!

PR #478, opened 2021-07-26
HOCS-3408 - Ability to send an initial response

agraham19:
* Move validity question from Draft to Creation
* Add template/upload doc screens


gary-newbolt (approved):
lgtm... 

Confirmed required that case Creation and Initial Response be separated post MVP

PR #479, opened 2021-07-26
HOCS-3394 MPAM PO Rejected cases return to QA Ministerial workstack

robertdeniszczyc2:
Send MPAM PO Rejected cases back to QA rather than Draft for UKVI/BF/IE


gardnersam:
@robertdeniszczyc2 Test coverage for MPAM_PO drops from  28.6% to 26.7% with this change, seeing as it's so low to begin with I think it would be worth at least maintaining the level of coverage we've got. Was surprised as it seems like you've added pretty good coverage in the unit test file. Suggests one of the new paths must be going untested but I can't quite tell which as the generated html files aren't colouring bits in for me.

gardnersam:
I was looking in the wrong folder for test coverage results, actual results are a good improvement

gardnersam (approved):

PR #486, opened 2021-08-12
HOCS-3618 FOI Allocation Note has null entry

LaurieBennett:
Remove null entry from end of allocation note for FOI Allocation step

### DIFF
--- a/src/main/java/uk/gov/digital/ho/hocs/workflow/BpmnService.java
+++ b/src/main/java/uk/gov/digital/ho/hocs/workflow/BpmnService.java
@@ -331,6 +331,18 @@ public class BpmnService {
         log.info("Adding Casenote to Case: {}", caseUUIDString);
     }
 
+    public String constructAllocationNote(String oldTeamName, String newTeamUUID, String stageName,
+                                          String allocationNote, String allocationNoteType){
+        String allocationNoteText = allocationNote.isBlank() ? " stage" :" stage: " + allocationNote;
+        if (allocationNoteType.equals("ALLOCATE")){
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/digital/ho/hocs/workflow/BpmnService.java
```diff
@@ -345,11 +345,13 @@ public void updateAllocationNoteWithDetails(String caseUUIDString, String stageU
         String oldTeamName =  infoClient.getTeam(caseworkClient.getStageTeam(UUID.fromString(caseUUIDString), UUID.fromString(allocationStageUUID))).getDisplayName();
         String caseworkStageType = caseworkClient.getStageType(UUID.fromString(caseUUIDString), UUID.fromString(allocationStageUUID));
         StageTypeDto stageTypeDto = infoClient.getAllStageTypes().stream().filter(st -> st.getType().equals(caseworkStageType)).findFirst().get();
+        String allocationNoteText = allocationNote.isBlank() ? " stage" :" stage: " + allocationNote;
```
gary-newbolt, line 348: Whilst this might solve the immediate issue... I don't think its a the right solution to the problem. I think you might be better off creating a new method in this class called `updateAllocationDetails` that constructs the string with the details only rather than if statement to manipulate the string in this way. It would be cleaner and more precise. You would need to alter the bpmn to call this new method. And write a test for your new method.

gary-newbolt:
@LaurieBennett your pipeline here has failed because one of the test in the workflow service has failed as a result of the change. Please make sure that before your next commit you run the test suit for the service to catch these.

LaurieBennett:
This has been merged in on another branch

PR #487, opened 2021-08-12
HOCS-3622 Split DCU MIN and DCU TRO Input screens

robertdeniszczyc2:
There are differences between the screens now so this makes them separate rather than sharing


gardnersam (approved):
Looks good

PR #488, opened 2021-08-13
HOCS-3262 Changes to final allocation screen

LaurieBennett:
1) Remove green "Case has been allocated to:..." panel from final screen in Allocation stage

2) Black text in screen shot to change from "Case FOI/xxxx... has been allocated to..." to "Case FOI/xxxx... will be allocated to..."

3) Green button to read "Confirm allocation"


gary-newbolt (approved):
lgtm

PR #492, opened 2021-08-18
HOCS-3647 add logic to draft bpmn diagram for back link

abarker66:
Added missing logic to draft stage bpmn diagram for back link.


LaurieBennett (approved):
Looks good to me, pass

PR #497, opened 2021-08-25
Feature/hocs 3481 foi sticky case

abarker66:
Added activity to FOI bpmn diagram to enable sticky case functionality. If a user is a member of the next team then the case is allocated automatically.


gary-newbolt (approved):
lgtm

PR #498, opened 2021-09-02
Feature/hocs 3558 exemptions and exceptions

LaurieBennett:
Added Dispatch step with exemptions and exceptions and tests for dispatch step with exemptions and exceptions


gary-newbolt:
Code wise it looks good however on the Workflow I think the `Dispatch Confirmation` is supposed to come after the capturing of the outcome details, I could be wrong though, if you've checked it and I'm wrong I'm happy to approve the code.

gary-newbolt (approved):
lgtm

PR #503, opened 2021-09-08
HOCS-3723 Added test and updated flow criterias for MPAM Campaign wor…

gardnersam:
…kflow

The MPAM campaigns workflow was untested, so tests are added here. The cause of the bug isn't 100% clear but other user input gates use the FORWARD and !FORWARD options for determining routes essentially building in a default route. I have updated the gate that the problem occurred to follow that pattern, functionality is uneffected but future errors should now be impossible.


robertdeniszczyc2 (approved):
Approving change as functionally it should cover the scenario and tests added to 100% coverage

PR #506, opened 2021-09-10
HOCS-3577 Merge hocs-workflow epic/foi-mvp branch into main

agraham19:
HOCS-3577 Merge hocs-workflow main branch into epic/foi-mvp


src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
```diff
@@ -66,9 +67,39 @@ public WorkflowService(CaseworkClient caseworkClient,
         this.userPermissionsService = userPermissionsService;
     }
 
-    public CreateCaseResponse createCase(String caseDataType, LocalDate dateReceived, List<DocumentSummary> documents, UUID userUUID, UUID fromCaseUUID) {
+    public CreateCaseResponse createCase(String caseDataType, LocalDate dateReceived, List<DocumentSummary> documents, UUID userUUID, UUID fromCaseUUID, Map<String, String> receivedData) {
```
gary-newbolt, line 70: Just to confirm, the front end call was changed to pass in the map of `receivedData` by default wasn't it? We should be careful on here.

gary-newbolt: I don't think we need to change anything, only that we make sure the corresponding change goes in to the `hocs-frontend` code.

src/main/java/uk/gov/digital/ho/hocs/workflow/client/caseworkclient/dto/AddressDto.java
```diff
@@ -6,7 +6,7 @@
 import lombok.Getter;
 import lombok.NoArgsConstructor;
 
-@AllArgsConstructor(access = AccessLevel.PRIVATE)
+@AllArgsConstructor()
```
gary-newbolt, line 9: Are we using this anywhere? Can the change be reverted? I can't seen an instance where we use the `new AddressDto` object instantiation anywhere... I might have missed it.

src/main/java/uk/gov/digital/ho/hocs/workflow/client/infoclient/dto/StageTypeDto.java
```diff
@@ -0,0 +1,38 @@
...
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+
+import java.io.Serializable;
+
+@NoArgsConstructor
+@Getter
+public class StageTypeDto implements Serializable {
+
+    //TODO: remove - used in UI
```
gary-newbolt, line 13: What would be the effort required to remove these now? Given they've been highlighted here?

gary-newbolt:
Generally looks okay to go... there are a few _possibly should do_ type changes but otherwise ok

LaurieBennett (approved):
All approved

PR #507, opened 2021-09-10
HOCS-3758 Capture Outcomes and Exemptions: Information released in full

agraham19:
* Skip to dispatch confirmation on full release


gary-newbolt (changes requested):
Line 102 in BPMN needs to have the gateway label changed to include `RELEASED_IN_FULL` event.

gary-newbolt (approved):

PR #508, opened 2021-09-10
HOCS-3767 Dispatch cases should be with DRAFT team not FOI allocation…

agraham19:
… team

* Return cases to draft team when a case is dispatched from approval


gary-newbolt (approved):
lgtm

PR #509, opened 2021-09-10
HOCS-3760 - Timeline: Allocation to Dispatch and Soft Close stages ap…

agraham19:
…pears twice

Before sticky cases, it was necessary to explicitly reassign the case to the
previous case owner at the dispatch and soft close stages. Now sticky cases 
is handling this, the case was getting assigned to the user twice, leading to
two audit messages.

* Remove superfluous user assignments at dispatch and soft close


gary-newbolt (approved):
All good!

PR #511, opened 2021-09-13
HOCS-3771 making selection of mpam clearance request status mandatory

gardnersam:
This PR alters workflow to use the new variable name for the radio buttons on the clearance fulfilment screen that was setup in the associated data PR


LiamMacP (approved):
Looks good to me...

PR #513, opened 2021-09-13
HOCS-3777: update rejected column value

LiamMacP:
Update the `Rejected` case data value to `By QA` when rejected from
the `QA (Secretariat Clearance Requested)` stage. This is then used in 
the workstack to identify where the case has been rejected from.


ARobinson26 (approved):
Looks good to me

PR #514, opened 2021-09-14
HOCS-3780 - Add diff based test for COMP bpmn files

fewtrees-ho:
Add a test to ensure we have checked and confirmed the differences between COMP & COMP2 BPMN files

### DIFF
--- a/src/test/scripts/check_COMP_COMP2.sh
+++ b/src/test/scripts/check_COMP_COMP2.sh
@@ -12,7 +12,7 @@ do
   rm -f diff.tmp
   echo "Checking $f $comp2file...\c"
   diff $f $comp2file > diff.tmp
-  diff diff.tmp $2/$diffImage
+  diff -b diff.tmp $2/$diffImage
   exitCode=$?
   rm -f diff.tmp
   if [ $exitCode -ne 0 ]; then
@@ -26,4 +26,4 @@ done
 
 # All good at this point
 exit 0
- 
+
--- a/src/test/scripts/check_COMP_COMP2.sh
+++ b/src/test/scripts/check_COMP_COMP2.sh
@@ -1,0 +1,29 @@
+# $1 - COMP file inc path
+# $2 - expected diffs directory
+
+for f in $1
+do
+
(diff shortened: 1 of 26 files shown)


src/test/scripts/diff_COMP_COMP2.sh
```diff
@@ -0,0 +1,36 @@
...
+  echo "             is then used in check_COMP_COMP2.sh as part of the Gradle test process."
+  echo "Usage:"
+  echo "COMP_COMP2_diff <source_file> <target_dir>"
+  echo "  <source_file> - The COMP bpmn file"
+  echo "  <target_dir>  = The target directory to store the diff file"
+}
+# $1 - COMP source file inc path
+# $2 - target dir path for output
+
+if [ -z "$1" ]; then
```
LiamMacP, line 14: Could be collapsed into one?

src/test/scripts/check_COMP_COMP2.sh
```diff
@@ -0,0 +1,29 @@
...
+  # Build the diff output filename
+  diffImage=`echo "$(basename $f)-$(basename $comp2file).diff" | sed 's/\.bpmn//g'`
+  rm -f diff.tmp
+  echo "Checking $f $comp2file...\c"
+  diff $f $comp2file > diff.tmp
+  diff diff.tmp $2/$diffImage
+  exitCode=$?
+  rm -f diff.tmp
+  if [ $exitCode -ne 0 ]; then
+     echo "Unexpected error found"
```
gardnersam, line 19: Could this print out what file it failed on?

fewtrees-ho:
It does as it's going through..

Each file is printed out and them appended with an 'OK' or details of the error 🙂
________________________________

src/test/scripts/check_COMP_COMP2.sh
gardnersam: Yep so it does, I should have looked 5 lines up!

gardnersam (approved):
Looks good, a good solution for enforcing keeping the stages in line

PR #524, opened 2021-10-01
HOCS-3894 Prevent parallel Docker push steps

zuzak:


gary-newbolt (commented):
On line 80, do we need a depends on condition with this one to avoid the parallel execution problem, i.e. `depends on: - build & push` ?

gary-newbolt (approved):

PR #531, opened 2021-10-12
Feature/hocs 3854 transfer smc to iedet

gardnersam:


src/main/resources/processes/SMC.bpmn
```diff
@@ -50,6 +50,7 @@
         <camunda:out source="CompType" target="CompType" />
         <camunda:out source="Stage" target="Stage" />
         <camunda:out source="ScmTriageResult" target="ScmTriageResult" />
+        <camunda:out source="CctTriageAccept" target="CctTriageAccept" />
```
robertdeniszczyc2, line 53: Is the SCM triage step different to the CCT triage step? I might need to see a demo of this to properly understand this flow

PR #533, opened 2021-10-14
HOCS-3970 Fix case creation for cases created by case-creator

robertdeniszczyc2:
Ensures case-creator can create cases and removes mutation of data to improve readability.

### DIFF
--- a/src/main/resources/processes/SMC.bpmn
+++ b/src/main/resources/processes/SMC.bpmn
@@ -50,6 +50,7 @@
         <camunda:out source="CompType" target="CompType" />
         <camunda:out source="Stage" target="Stage" />
         <camunda:out source="ScmTriageResult" target="ScmTriageResult" />
+        <camunda:out source="CctTriageAccept" target="CctTriageAccept" />
       </bpmn:extensionElements>
       <bpmn:incoming>Flow_08fga3p</bpmn:incoming>
       <bpmn:outgoing>Flow_115y0ny</bpmn:outgoing>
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
```diff
@@ -69,53 +69,48 @@ public WorkflowService(CaseworkClient caseworkClient,
 
     public CreateCaseResponse createCase(String caseDataType, LocalDate dateReceived, List<DocumentSummary> documents, UUID userUUID, UUID fromCaseUUID, Map<String, String> receivedData) {
         // Create a case in the casework service in order to get a reference back to display to the user.
-        Map<String, String> data = new HashMap<>(receivedData);
         CreateCaseworkCorrespondentRequest correspondentRequest = null;
```
LiamMacP, line 72: Defaults to `null` so not needed.

src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
```diff
@@ -69,53 +69,48 @@ public WorkflowService(CaseworkClient caseworkClient,
...
-                    .address1(data.get("Address1"))
-                    .address2(data.get("Address2"))
-                    .address3(data.get("Address3"))
-                    .country(data.get("Country"))
-                    .telephone(data.get("Telephone"))
-                    .email(data.get("Email"))
-                    .reference(data.get("Reference"))
-                    .build();
-
-            data.remove("Fullname");
```
LiamMacP, line 91: Wow, glad to see this going 👍🏽

src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
robertdeniszczyc2: If the assign to null is removed the null check on line 115 warns that the variable might not have been initialised.

robertdeniszczyc2:
@LiamMacP are you happy with me Resolving the comments added please? Cheers

LiamMacP:
Yep fine with me, something to revisit in the future...

src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowConstants.java
```diff
@@ -7,4 +7,20 @@
     String LAST_UPDATED_BY_USER = "LastUpdatedByUserUUID";
     String DATE_RECEIVED = "DateReceived";
     String CASE_REFERENCE = "CaseReference";
+    String FULL_NAME = "Fullname";
```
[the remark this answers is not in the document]
robertdeniszczyc2: final is redundant on interface fields so not required

PR #545, opened 2021-11-03
HOCS-3987: Add pending path to SMC_TRIAGE

JosephRogerson:
Currently, there isn't a path for pending in the SMC_TRIAGE model. This
means that the case moves to draft instead of staying in triage. This
change adds this path and the required test coverage.


LiamMacP (approved):
Looks good, I'll show you next time how to make these slightly more readable for peer reviews! good work!

PR #547, opened 2021-11-05
HOCS-4030 move the kube config into the main project

nickmcmahon01:


LiamMacP (changes requested):
This still needs to have the history brought across before we can merge 🙂

PR #552, opened 2021-11-12
HOCS-4081 add check for null field when adding case notes

abarker66:
added check for null field when adding case notes. Null field is replaced by empty string and error logged as this should not really be happening.


gary-newbolt (approved):
lgtm

PR #556, opened 2021-11-23
Feature/hocs 4023 close case endpoint

gardnersam:
This PR adds an endpoint to workflow which will close the case linked to the given UUID. The intention behind the implementation is to fully revert any changes if an error occurs up to the final step of deleting the Camunda process.

### DIFF
--- a/src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
+++ b/src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
@@ -15,6 +15,7 @@ import uk.gov.digital.ho.hocs.workflow.client.caseworkclient.CaseworkClient;
 import uk.gov.digital.ho.hocs.workflow.client.caseworkclient.dto.CreateCaseworkCaseResponse;
 import uk.gov.digital.ho.hocs.workflow.client.caseworkclient.dto.GetAllStagesForCaseResponse;
 import uk.gov.digital.ho.hocs.workflow.client.caseworkclient.dto.GetCaseworkCaseDataResponse;
+import uk.gov.digital.ho.hocs.workflow.client.caseworkclient.dto.GetStagesResponse;
 import uk.gov.digital.ho.hocs.workflow.client.caseworkclient.dto.StageDto;
 import uk.gov.digital.ho.hocs.workflow.client.documentclient.DocumentClient;
 import uk.gov.digital.ho.hocs.workflow.client.infoclient.InfoClient;
@@ -366,8 +367,8 @@ public class WorkflowService {
         GetCaseworkCaseDataResponse caseDetails = caseworkClient.getCase(caseUUID);
 
         //Get stage uuid from casework
-        String stage = caseworkClient.getActiveStage(caseDetails.getReference().replace("/", "%2F"));
-        UUID stageUUID = UUID.fromString(getStageUUID(stage));
+        GetStagesResponse stage = caseworkClient.getActiveStage(caseDetails.getReference().replace("/", "%2F"));
+        UUID stageUUID = stage.getStages().stream().findFirst().get().getStageUUID();
 
         //Mark case as complete
         caseworkClient.completeCase(caseUUID, true);
--- a/src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
+++ b/src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
@@ -393,12 +393,6 @@ public class WorkflowService {
         return ResponseEntity.ok("Closed case " + caseUUID);
     }
 
-    private String getStageUUID(String stage) throws JsonProcessingException {
-        ObjectMapper objectMapper = new ObjectMapper();
-        JsonNode node = objectMapper.readTree(stage).get("stages").get(0).get("uuid");
-        return node.toString().substring(1, node.toString().length() -1);
-    }
-
     public void deleteProcess(UUID stageUuid){
         camundaClient.removeProcess(stageUuid);
     }
--- a/src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
+++ b/src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
@@ -1,8 +1,12 @@
 package uk.gov.digital.ho.hocs.workflow.api;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import lombok.extern.slf4j.Slf4j;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
 import org.springframework.stereotype.Service;
 import org.springframework.web.client.HttpClientErrorException;
 import uk.gov.digital.ho.hocs.workflow.api.dto.*;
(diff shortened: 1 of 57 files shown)


src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
```diff
@@ -357,4 +361,45 @@ public void updateStage(UUID caseUUID, UUID stageUUID, Map<String, String> value
...
+
+        //Get stage uuid from casework
+        String stage = caseworkClient.getActiveStage(caseDetails.getReference().replace("/", "%2F"));
+        UUID stageUUID = UUID.fromString(getStageUUID(stage));
+
+        //Mark case as complete
+        caseworkClient.completeCase(caseUUID, true);
+
+        //Update the stage team to null
+        UUID oldTeam = caseworkClient.getStageTeam(caseUUID, stageUUID); //To allow reversion if required
```
nickmcmahon01, line 376: Not a blocking comment, but should the 'completeCase' method in Casework->CaseService perform this action instead? you could put that behaviour in a transaction with the complete=true and not have to do the countermeasure.

src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
```diff
@@ -357,4 +361,45 @@ public void updateStage(UUID caseUUID, UUID stageUUID, Map<String, String> value
 
         camundaClient.completeTask(stageUUID, values);
     }
+
+    public ResponseEntity closeCase(UUID caseUUID) throws JsonProcessingException {
+        GetCaseworkCaseDataResponse caseDetails = caseworkClient.getCase(caseUUID);
+
+        //Get stage uuid from casework
+        String stage = caseworkClient.getActiveStage(caseDetails.getReference().replace("/", "%2F"));
```
LiamMacP, line 369: Unsure what the casework endpoint returns but mapping to a DTO would be preferred? I suspect there should only be 1 active stage for a case?

src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
```diff
@@ -357,4 +361,45 @@ public void updateStage(UUID caseUUID, UUID stageUUID, Map<String, String> value
...
+        UUID stageUUID = UUID.fromString(getStageUUID(stage));
+
+        //Mark case as complete
+        caseworkClient.completeCase(caseUUID, true);
```
LiamMacP, line 373: What if this fails? Would we not want this to undo?

src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
nickmcmahon01: its a Set<> but there should only be one - the set was for parallel stages.

src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
gardnersam: If this call fails then nothing will have happened so far so there shouldn't be anything to undo.

src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
gardnersam: I've swapped this to a DTO, will make it more useful in the future if people want to get more out of it

src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
LiamMacP: Yeah looked at the `completeCase` method, seems relatively safe...

src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
```diff
@@ -357,4 +362,45 @@ public void updateStage(UUID caseUUID, UUID stageUUID, Map<String, String> value
...
+            caseworkClient.updateStageTeam(caseUUID, stageUUID, oldTeam, null);
+            caseworkClient.completeCase(caseUUID, caseDetails.getCompleted()); //Audit event may already have been created for closed case
+            return ResponseEntity.status(HttpStatus.INTERNAL_SERVER_ERROR).body("Failed to delete process for " + caseUUID + " Error : " + e);
+        }
+        return ResponseEntity.ok("Closed case " + caseUUID);
+    }
+
+    private String getStageUUID(String stage) throws JsonProcessingException {
+        ObjectMapper objectMapper = new ObjectMapper();
+        JsonNode node = objectMapper.readTree(stage).get("stages").get(0).get("uuid");
```
LiamMacP, line 398: Would mapping that into a object be simpler to read?

src/main/java/uk/gov/digital/ho/hocs/workflow/client/caseworkclient/dto/GetStagesResponse.java
```diff
@@ -0,0 +1,29 @@
...
+
+@AllArgsConstructor
+@NoArgsConstructor
+@Getter
+public class GetStagesResponse {
+
+    @JsonProperty("stages")
+    private Set<GetStageResponseDto> stages;
+
+//    public static GetStagesResponse from(Set<StageDto> stages) {
```
LiamMacP, line 21: Commented out?

src/test/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowServiceTest.java
```diff
@@ -7,21 +7,16 @@
...
 import static org.mockito.Mockito.when;
 
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.LinkedList;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.UUID;
+import java.util.*;
```
LiamMacP, line 10: Not an blocking issue, but bringing in the entire util and mockito packages... there is a IntelliJ setting that disables this.

src/test/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowServiceTest.java
```diff
@@ -706,4 +705,55 @@ private SchemaDto exampleSchemaDto() {
...
+        getCaseworkCaseDataResponse.setCompleted(false);
+        Mockito.doReturn(getCaseworkCaseDataResponse).when(caseworkClient).getCase(any());
+
+        GetStageResponseDto getStageResponseDto = new GetStageResponseDto(UUID.fromString(stageUUID), null, null);
+        Set<GetStageResponseDto> stages = new HashSet<>(Collections.singletonList(getStageResponseDto));
+        GetStagesResponse getStagesResponse = new GetStagesResponse(stages);
+
+        Mockito.doReturn(getStagesResponse).when(caseworkClient).getActiveStage(anyString());
+
+//        Mockito.doReturn("{\"stages\" : [ { \"uuid\" : \"" + stageUUID + "\" } ]}").when(caseworkClient).getActiveStage(anyString());
```
LiamMacP, line 753: Commented out, does this need removing?

src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
gardnersam: This method was no longer used after getting a DTO rather than a string in an earlier review comment so I've removed it

src/main/java/uk/gov/digital/ho/hocs/workflow/client/caseworkclient/dto/GetStagesResponse.java
gardnersam: Woops, removed

src/test/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowServiceTest.java
gardnersam: Fair, I've updated to single imports

src/test/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowServiceTest.java
gardnersam: 2nd woops, removed

src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
gardnersam: We'd need the Complete status and current team anyway in order to reset if killing the camunda process failed so I think this is still the way to go

src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
```diff
@@ -357,4 +362,39 @@ public void updateStage(UUID caseUUID, UUID stageUUID, Map<String, String> value
 
         camundaClient.completeTask(stageUUID, values);
     }
+
+    public ResponseEntity closeCase(UUID caseUUID) throws JsonProcessingException {
+        GetCaseworkCaseDataResponse caseDetails = caseworkClient.getCase(caseUUID);
+
+        //Get stage uuid from casework
+        GetStagesResponse stage = caseworkClient.getActiveStage(caseDetails.getReference().replace("/", "%2F"));
```
fewtrees-ho, line 370: I think there's a URLEncode method lurking around that would do this properly..

fewtrees-ho (changes requested):

src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
gardnersam: https://github.com/UKHomeOffice/hocs-frontend/blob/main/server/libs/encodingHelpers.js - we have this in the frontend but can't find anything similar in workflow.

fewtrees-ho: https://docs.oracle.com/javase/7/docs/api/java/net/URLEncoder.html

fewtrees-ho (approved):

PR #559, opened 2021-11-25
HOCS-4169: create skeleton workflow caseType TO

gary-newbolt:


abarker66 (approved):
Looks good, I think we may need a key on the diagram to explain the psychedelic colour scheme.
The transferred cases are just trapped at triage it seems - is there an escape/stop route for these cases once transferred or removed?

PR #563, opened 2021-11-29
HOCS-4259: increase logging around createStage team/user assignment

gary-newbolt:


src/main/java/uk/gov/digital/ho/hocs/workflow/BpmnService.java
```diff
@@ -509,25 +509,26 @@ public void createCaseConversionNote(String caseUUIDString, String stageUUIDStri
 
     private UUID deriveTeamUUID(String caseUUIDString, String stageTypeString, String allocationTeamString) {
         UUID teamUUID;
-        if (StringUtils.isEmpty(allocationTeamString)) {
+        if (!StringUtils.hasText(allocationTeamString)) {
```
LiamMacP, line 512: Is it worth logging out this allocationTeamString as well?

src/main/java/uk/gov/digital/ho/hocs/workflow/BpmnService.java
```diff
@@ -509,25 +509,26 @@ public void createCaseConversionNote(String caseUUIDString, String stageUUIDStri
...
         return teamUUID;
 
     }
 
     private UUID deriveUserUUID(String caseUUIDString, String stageTypeString, String allocatedUserId) {
-        UUID userUUID = null;
-        if (!StringUtils.isEmpty(allocatedUserId)) {
-            log.debug("Assigning user {} to stage {} for case {}", allocatedUserId, stageTypeString, caseUUIDString);
-            userUUID = UUID.fromString(allocatedUserId);
+        if (StringUtils.hasText(allocatedUserId)) {
```
LiamMacP, line 525: Same here, worth logging out the allocatedUserId?

src/main/java/uk/gov/digital/ho/hocs/workflow/BpmnService.java
gary-newbolt: In this one... that if block is only called if the `allocatedTeamString` is `null` anyway and in both the subsequent log output it logs the resolved teamUUID, specifically the new final info level one. From the logic there's no way if can escape that method as null, if the `infoClient` call fails it should trigger the global handling anyway.

src/main/java/uk/gov/digital/ho/hocs/workflow/BpmnService.java
gary-newbolt: I think the same is true here... if there is an `allocatedUserId` it's logged out in the next line. If there isn't, there's a `warn` level message explicitly stating it.

LiamMacP (approved):

PR #564, opened 2021-12-02
HOCS-4238: Update CompType when transferring case

JosephRogerson:
At the moment when a case is transferred via CCH returns the CompType is
not changed. This change adds an activity to the COMP and COMP2 CCH flow
that updates the CompType before transfer.


fewtrees-ho (changes requested):
Hi Joseph,

There are no unit tests for this change. Would you be able to add them in?

Cheers
Rob

JosephRogerson:
> Hi Joseph,
> 
> Should the value change happen in the backend? Is it a backend activity?
> 
> There are no unit tests for this change. Would you be able to add them in?
> 
> Cheers Rob

Hi Rob,

The value change will happen in the backend, currently when the team is changed in the CCH returns process the CompType is not updated to match. This uses the `updateValue` method to set this value.

I'll add some unit tests for this.

Thanks,
Joseph

fewtrees-ho (approved):

PR #566, opened 2021-12-02
HOCS-4297-Add-jacoco-and-sonar

fewtrees-ho:


nickmcmahon01 (approved):

build.gradle
```diff
@@ -108,6 +108,26 @@ configurations {
     integrationTestRuntimeOnly.extendsFrom testRuntime
 }
 
+plugins.withType(JacocoPlugin) {
+    tasks["test"].finalizedBy 'jacocoTestReport'
+}
+
+test {
+    maxHeapSize = "1024m"
```
nickmcmahon01, line 116: Not sure what memory we have available on drone... Let's see what happens

.drone.yml
```diff
@@ -36,7 +36,7 @@ steps:
   - name: sonar scanner
     image: quay.io/ukhomeofficedigital/sonar-scanner:v3.0.2
     depends_on:
-      - build project
+      - test project
```
zuzak, line 39: This changes it to run in serial to the test step instead of parallel with it -- is that intentional?

PR #567, opened 2021-12-03
HOCS-4121 added data input stage for treat official

abarker66:
This delivers the data input stage for treat official cases.


gary-newbolt (approved):
lgtm - I agree with you that the team uuid in there is a bit ugh! but it does the job.

PR #569, opened 2021-12-03
HOCS-4312 - Remove the Retry annotation

nickmcmahon01:
As part of the investigation it was determined that the application was failing to create new threads for the retry code, we have given the pods more memory but we don't believe we need to use @Retry at all - we prefer to fail fast.


robertdeniszczyc2 (approved):
Talked through changes with Nick and approve

PR #580, opened 2021-12-14
Feature/hocs 4327 changes to recipient capture

abarker66:
Added changes to data input stage to allow a user to skip adding a recipient to the case.


gary-newbolt (approved):
lgtm

PR #595, opened 2021-12-22
📌 HOCS-4212 - Upgrade Postgres version used in the project

nickmcmahon01:
At the moment we are using unknown versions of Postgres in the build, this commit moves drone and docker-compose over to using the official images.


LiamMacP (approved):
Change looks good to me...

PR #602, opened 2022-01-13
HOCS-4134 Campaigns

agraham19:
* Allow TO cases to be moved to a campaign

### DIFF
--- a/src/main/resources/processes/TO_QA.bpmn
+++ b/src/main/resources/processes/TO_QA.bpmn
@@ -8,7 +8,6 @@
       <bpmn:incoming>Flow_0rvjo1x</bpmn:incoming>
       <bpmn:incoming>Flow_15zn9g3</bpmn:incoming>
       <bpmn:incoming>Flow_1tgyxfj</bpmn:incoming>
-      <bpmn:incoming>Flow_1sjleue</bpmn:incoming>
     </bpmn:endEvent>
     <bpmn:sequenceFlow id="Flow_0flmj05" sourceRef="StartEvent_1" targetRef="TO_QA_OUTCOME" />
     <bpmn:sequenceFlow id="Flow_0i8bf5b" sourceRef="TO_QA_OUTCOME" targetRef="Gateway_0vour6f" />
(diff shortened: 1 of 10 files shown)


gary-newbolt (changes requested):
When you start the service the change has introduced a new warning messages:
`2022-01-14 08:59:37.325  WARN 83994 --- [           main] org.camunda.bpm.engine.util              : ENGINE-09004 Warnings during parsing: Exclusive Gateway 'Gateway_1w45vlo' has outgoing sequence flow 'Flow_07sgk52' without condition which is not the default flow. We assume it to be the default flow, but it is bad modeling practice, better set the default flow in your gateway. | /Users/garynewbolt/Development/HO/hocs-workflow/build/resources/main/processes/TO_TRIAGE.bpmn | line 0 | column 0`

Can that be fixed please?

src/main/resources/processes/TO_QA.bpmn
```diff
@@ -75,116 +78,160 @@
...
+    <bpmn:exclusiveGateway id="Gateway_05qknzm" name="Direction?" default="Flow_1sjleue">
+      <bpmn:incoming>Flow_1j2kqxk</bpmn:incoming>
+      <bpmn:outgoing>Flow_1ekh7v9</bpmn:outgoing>
+      <bpmn:outgoing>Flow_1sjleue</bpmn:outgoing>
+    </bpmn:exclusiveGateway>
+    <bpmn:sequenceFlow id="Flow_1j2kqxk" sourceRef="TO_GET_CAMPAIGN_TYPE" targetRef="Gateway_05qknzm" />
+    <bpmn:sequenceFlow id="Flow_1ekh7v9" sourceRef="Gateway_05qknzm" targetRef="TO_QA_OUTCOME">
+      <bpmn:conditionExpression xsi:type="bpmn:tFormalExpression">${ DIRECTION == "BACKWARD" }</bpmn:conditionExpression>
+    </bpmn:sequenceFlow>
+    <bpmn:sequenceFlow id="Flow_1sjleue" sourceRef="Gateway_05qknzm" targetRef="Event_0nlva02" />
```
gary-newbolt, line 97: This flow also should not be going straight to the end, rather to the `Update Approval: DecisionNotYetMade` process task.

gary-newbolt (approved):

PR #607, opened 2022-01-20
HOCS-4454 Change business area link broken at QA stage

agraham19:
* add required workflow paths for QA change business area


gary-newbolt (approved):
lgtm

PR #612, opened 2022-01-24
Create codeql-analysis.yml

nickmcmahon01:


LiamMacP (approved):
Change looks good - approved.

PR #613, opened 2022-01-25
HOCS-4440: WebForm complaints allows early closure

LiamMacP:
This changes support the requirement for a user to be able to close a
WebForm case as early as possible if it's not deemed valid. This is done
through a check if the channel is specified as WebForm and then shows a
validation screen within COMP_REGISTRATION. The outcome of this screen
is then propagated up to the COMP workflow to complete the case as
required.

WIP change for closing Webform cases earlier in the pipeline. This
contains the core logic that works - the other parts (form design)
should be easily alterable).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -162,8 +162,8 @@ task checkCOMPSnapshots {
         exec {
             commandLine 'src/test/scripts/check_COMP_COMP2.sh', compFiles, compDiffs
         }
-
     }
+
 }
 
 jar {
--- a/src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
+++ b/src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
@@ -102,19 +102,19 @@ public class WorkflowService {
 
             // Start a new camunda workflow (caseUUID is the business key).
             Map<String, String> seedData = new HashMap<>();
-            if (Objects.equals(caseDataType, WorkflowConstants.CASE_DATA_TYPE_FOI) && receivedData != null) {
-                seedData.put(WorkflowConstants.CASE_REFERENCE, caseResponse.getReference());
-                seedData.put(WorkflowConstants.DATE_RECEIVED, dateReceived.toString());
-                seedData.put(WorkflowConstants.LAST_UPDATED_BY_USER, userUUID.toString());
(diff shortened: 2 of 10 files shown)


build.gradle
```diff
@@ -149,21 +149,7 @@ task createCOMPSnapshots {
 
 task checkCOMPSnapshots {
 
-    doFirst {
```
robertdeniszczyc2, line 152: Is this removal intentional?

src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
```diff
@@ -103,18 +103,18 @@ public CreateCaseResponse createCase(String caseDataType, LocalDate dateReceived
...
                 seedData.put(WorkflowConstants.TOPICS, receivedData.get(WorkflowConstants.TOPICS));
                 seedData.put(WorkflowConstants.REQUEST_QUESTION, receivedData.get(WorkflowConstants.REQUEST_QUESTION));
                 seedData.put(WorkflowConstants.ORIGINAL_CHANNEL, receivedData.get(WorkflowConstants.ORIGINAL_CHANNEL));
-            } else {
-                seedData.put(WorkflowConstants.CASE_REFERENCE, caseResponse.getReference());
-                seedData.put(WorkflowConstants.DATE_RECEIVED, dateReceived.toString());
-                seedData.put(WorkflowConstants.LAST_UPDATED_BY_USER, userUUID.toString());
+            } else if (Objects.equals(caseDataType, WorkflowConstants.CASE_DATA_TYPE_COMP)
+                    && receivedData != null
+                    && Objects.equals(receivedData.get(WorkflowConstants.CHANNEL), WorkflowConstants.CHANNEL_COMP_WEBFORM)) {
```
robertdeniszczyc2, line 112: Is it worth extracting this logic chain into a method with a more readable name so it's easier to understand this block? Would also be able to add test for the logic chain easily then too.

robertdeniszczyc2 (changes requested):
Added a couple questions,

Cheers,
Rob

build.gradle
LiamMacP: Not intentional - was just for initial testing - added back

src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
LiamMacP: Good shout simplified it in this area - and added a test

robertdeniszczyc2 (approved):

PR #614, opened 2022-01-25
Feature/hocs 4135 put case in stop list

abarker66:
Added stop list to treat Official workflow.


src/main/resources/processes/TO.bpmn
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
```
gary-newbolt, line 1: The change business area functionality requires the stage on this diagram to support transferring it to the different business area, on the `TO.bpmn` it's a gateway labelled `Transfer Case?`

gary-newbolt (approved):
lgtm

PR #616, opened 2022-01-28
HOCS-4462: alter driving variable for WebForm case

LiamMacP:
To better support strictly WebForm cases for COMP in the future, 
the driving process variable has been altered to the non-modifiable
variable called `XOriginatedFrom`. 

It also supports the passing through the received data to the case 
data if it originates from the UKVI WebForm.

### DIFF
--- a/src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
+++ b/src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
@@ -81,15 +81,18 @@ public class WorkflowService {
         caseData.put(WorkflowConstants.DATE_RECEIVED, dateReceived.toString());
         caseData.put(WorkflowConstants.LAST_UPDATED_BY_USER, userUUID.toString());
 
-        if (receivedData != null && Objects.equals(caseDataType, WorkflowConstants.CASE_DATA_TYPE_FOI)) {
-            // If we have a name we have correspondent details attached to the case creation request
-            if(receivedData.containsKey(WorkflowConstants.FULL_NAME)) {
+        if (receivedData != null) {
+            // If FOI and we have a name we have correspondent details attached to the case creation reques
+            if (Objects.equals(caseDataType, WorkflowConstants.CASE_DATA_TYPE_FOI)
+                    && receivedData.containsKey(WorkflowConstants.FULL_NAME)) {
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
```diff
@@ -81,15 +81,18 @@ public CreateCaseResponse createCase(String caseDataType, LocalDate dateReceived
         caseData.put(WorkflowConstants.DATE_RECEIVED, dateReceived.toString());
         caseData.put(WorkflowConstants.LAST_UPDATED_BY_USER, userUUID.toString());
 
-        if (receivedData != null && Objects.equals(caseDataType, WorkflowConstants.CASE_DATA_TYPE_FOI)) {
-            // If we have a name we have correspondent details attached to the case creation request
-            if(receivedData.containsKey(WorkflowConstants.FULL_NAME)) {
+        if (receivedData != null) {
+            // If FOI and we have a name we have correspondent details attached to the case creation reques
```
robertdeniszczyc2, line 85:
```suggestion
            // If FOI and we have a name we have correspondent details attached to the case creation request
```

robertdeniszczyc2:
Is the field `XOriginatedFrom` non-modifiable only by convention, i.e. it is technically modifiable but we indicate with the 'X' as a marker to signify that it shouldn't be modified? If so wonder if would be better with something more explicit, see this in React: https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml

Whichever option we do it should be documented in a README of the module it is appropriate, probably this one maybe under a Camunda Conventions section or something.

LiamMacP:
> Is the field `XOriginatedFrom` non-modifiable only by convention, i.e. it is technically modifiable but we indicate with the 'X' as a marker to signify that it shouldn't be modified? If so wonder if would be better with something more explicit, see this in React: https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml
> 
> Whichever option we do it should be documented in a README of the module it is appropriate, probably this one maybe under a Camunda Conventions section or something.

The README file has been cleaned up to include relevant information and now contains a subsection for `Camunda Conventions`.

robertdeniszczyc2 (approved):

PR #618, opened 2022-01-31
HOCS-4513 Changing primary correspondent does not update the summary …

agraham19:
…page for Treat Official

* Add step to update case primary correspondent


LaurieBennett (approved):
LGTM pass

PR #625, opened 2022-02-08
Hocs 4206 add home sec interest

LaurieBennett:


abarker66:
HI Laurie, you need to merge (manually) the four bpmn diagrams - the tests fail and the workflow is broken until this is resolved.

abarker66:
Hi Laurie, the build is still failing because the top level bpmn process tests all fail. I think I can see why; the HomeSec stage on the top level bpmn diagram is floating, as in not connected to anything!?! I think the line that looks like it is connected actually passes underneath it!

abarker66 (approved):
All looks good, tests pass.

PR #626, opened 2022-02-08
HOCS-4534 Close case without having to complete process

agraham19:
* Add close option to Triage, Draft, and QA


LaurieBennett (approved):
LGTM

PR #628, opened 2022-02-09
HOCS-4431-change code to not make calls to info service in stage crea…

gary-newbolt:
…tion


LiamMacP (approved):
Approved - change looks fine - confirmed code was moved to casework.

PR #636, opened 2022-02-18
HOCS-4578 Case data type request param

robertdeniszczyc2:
Adds RequestParam for type when updating case data

### DIFF
--- a/src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
+++ b/src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
@@ -412,8 +412,9 @@ public class WorkflowService {
         caseworkClient.updateCase(caseUUID, stageUUID, request);
 
         if (caseDataType != null) {
+            // Check here that the type is supported and in future this would allow for custom note text per type update
             if ("EX_GRATIA_UPDATE".equals(caseDataType)) {
-                caseworkClient.createCaseNote(caseUUID, caseDataType, "Updated Ex-Gratia case data");
+                caseworkClient.createCaseNote(caseUUID, caseDataType, "Case data updated");
             } else {
                 log.warn("Unable to create case note for case data update of type {}", caseDataType, value(EVENT, CASE_NOTE_FAILED));
             }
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowService.java
```diff
@@ -405,10 +406,19 @@ public void deleteProcess(UUID stageUuid){
...
     }
 
-    public void updateCaseDataValues(UUID caseUUID, UUID stageUUID, Map<String, String> request) {
+    public void updateCaseDataValues(UUID caseUUID, UUID stageUUID, String caseDataType, Map<String, String> request) {
         log.debug("Updating case data for case {} with stage {}", caseUUID, stageUUID);
         camundaClient.updateTask(stageUUID, request);
         caseworkClient.updateCase(caseUUID, stageUUID, request);
+
+        if (caseDataType != null) {
+            if ("EX_GRATIA_UPDATE".equals(caseDataType)) {
```
gary-newbolt, line 415: @robertdeniszczyc2 is there anyway this could be lifted out of here? It feels like this is too specific for a specific use case rather than a generic method. Maybe attach the note to the quest body?

robertdeniszczyc2: I agree it's not great. I can't pass it in the request body as the entire body is passed into the updateCase() and updateTask methods on line 411/412, and I don't want to process the case note text as a case data value. The request body only contains key value pairs of case data values for the case data blob, so a case note would pollute that data.

I can take another look though

LiamMacP: Its a fairly big piece but I was wondering if we could have a base WorkflowService that is derived from specific case type ones for specific functionality.

gary-newbolt: Another suggestion to keep it clean is to use the `NoteType` enum and set a field on the enum of `defaultNote` or something like that, then have the text come from the default note?

robertdeniszczyc2: I could set the type on the NoteType enum and here check if the type exists. For the text rather than being specific I go could generic like "Case data updated". That could work as the specific text doesn't add anything - we already render the title in the timeline as "Ex-Gratia Updated" so we're just repeating what's already been said with specific text here.

gary-newbolt: I think either would be fine. In my mind, I think it better that if submitted with an "EX_GRATIA_UPDATE" we might as well be specific in the message, and the `NoteType` hold a default, and another enum val of "CASE_DATA_UPDATE", with it's own default later if we need it at a later date. 

Or as you say... no note on it... it would be good to test how that looks and if it's ok with the just the title on the note, go for it. I think it would be good to have an overloaded method that passes `null` to this one would be good so the developer can choose either, but not beholden to that.

robertdeniszczyc2: Updated to have case note text specific to Ex-Gratia and a default. I recall recently there was a change to prevent creating case notes with blank/null content so I don't want to breach that rule here.

gary-newbolt (approved):
lgtm

PR #639, opened 2022-02-23
Hocs 4533 add return case to cch

LaurieBennett:


src/test/java/uk/gov/digital/ho/hocs/workflow/processes/to/TO.java
```diff
@@ -204,6 +209,146 @@ public void testAllCloseAfterTriage() {
...
+
+        verify(TOProcess, times(0))
+                .hasCompleted(DISPATCH);
+
+        // CASE COMPLETED
+        verify(bpmnService).completeCase(any());
+    }
+
+    @Test
+    public void testTransferFromCCHTriage(){
```
gary-newbolt, line 277: @LaurieBennett can you add tests for transfer at Draft and QA please? I don't mind if you combine them with this, or have separate tests, I'd just like them covered.

gary-newbolt:
On start up of the workflow service there's 3 warnings that have appeared related to the TO workflow:

`* Exclusive Gateway 'Gateway_1ykmpa4' has outgoing sequence flow 'Flow_1lvfgw9' without condition which is not the default flow. We assume it to be the default flow, but it is bad modeling practice, better set the default flow in your gateway. | /Users/garynewbolt/Development/HO/hocs-workflow/build/resources/main/processes/TO_HOME_SEC.bpmn | line 0 | column 0
2022-02-24 14:40:48.484  WARN 62194 --- [           main] org.camunda.bpm.engine.util              : ENGINE-09004 Warnings during parsing: 
* Exclusive Gateway 'Gateway_0d7zij1' has outgoing sequence flow 'Flow_0ywlba1' without condition which is not the default flow. We assume it to be the default flow, but it is bad modeling practice, better set the default flow in your gateway. | /Users/garynewbolt/Development/HO/hocs-workflow/build/resources/main/processes/TO.bpmn | line 0 | column 0
* Exclusive Gateway 'Gateway_00zox1a' has outgoing sequence flow 'Flow_0jbdluc' without condition which is not the default flow. We assume it to be the default flow, but it is bad modeling practice, better set the default flow in your gateway. | /Users/garynewbolt/Development/HO/hocs-workflow/build/resources/main/processes/TO.bpmn | line 0 | column 0
` 
Can you look into and resolve them please?

src/test/java/uk/gov/digital/ho/hocs/workflow/processes/to/TO.java
LaurieBennett: Added

gary-newbolt (approved):
lgtm

PR #642, opened 2022-02-24
HOCS-4602: Case migration

JosephRogerson:


src/main/java/uk/gov/digital/ho/hocs/workflow/BpmnService.java
```diff
@@ -103,6 +104,20 @@ public void createCase(@NotNull String caseType, @NotNull String dateReceived, @
         }
     }
 
+    public void migrateCase(@NotNull String caseType, @NotNull String dateReceived, @NotNull String fromCaseUUID) {
```
robertdeniszczyc2, line 107: Is caseType the source or target case type of the migration?

JosephRogerson: It's the target case type

src/main/java/uk/gov/digital/ho/hocs/workflow/BpmnService.java
```diff
@@ -103,6 +104,20 @@ public void createCase(@NotNull String caseType, @NotNull String dateReceived, @
...
+        UUID caseUuid = UUID.fromString(fromCaseUUID);
+        GetCaseworkCaseDataResponse caseData = caseworkClient.getCase(caseUuid);
```
gary-newbolt, line 109: I'm thinking on reflection that the logic to create the new Case of the new case type, might be better suited to take place in the `Casework Service` making this a very quick and easy pass through method. That way you could avoid making several Rest calls for the process. 

Just one migrate call, passing along the caseUUID, stageUUID, and the new casetype. What do you think?

We could also eliminate all of this data mapping here and in the `WorkflowService` class

src/main/java/uk/gov/digital/ho/hocs/workflow/client/caseworkclient/CaseworkClient.java
```diff
@@ -40,6 +40,13 @@ public CreateCaseworkCaseResponse createCase(String caseDataType, Map<String, St
         return response;
     }
 
+    public MigrateCaseworkCaseResponse migrateCase(String caseDataType, Map<String, String> data, LocalDate dateReceived, UUID fromCaseUUID) {
+        MigrateCaseworkCaseRequest request = new MigrateCaseworkCaseRequest(caseDataType, data, dateReceived, fromCaseUUID);
+        MigrateCaseworkCaseResponse response = restHelper.post(serviceBaseURL, "/migrate", request, MigrateCaseworkCaseResponse.class);
```
gary-newbolt, line 45: It might seem irrelevant, but it would be more consistent for the receiving api to have this URI: `/case/{fromCaseId}/migrate`

src/test/java/uk/gov/digital/ho/hocs/workflow/api/dto/MigrateCaseResponseTest.java
```diff
@@ -0,0 +1,33 @@
...
+
+        assertThat(migrateCaseResponse.getUuid()).isEqualTo(caseUUID);
+        assertThat(migrateCaseResponse.getReference()).isEqualTo(reference);
+
+    }
+
+    @Test
+    public void getMigrateCaseResponseNull() {
+
+        MigrateCaseResponse migrateCaseResponse = new MigrateCaseResponse(null, null);
```
gary-newbolt, line 27: If you're creating a request to migrate a case, and your input is `null`, `null`, would you be happen with a positive response? Which would, presumably continue to try to migrate case with ID `null`?

src/test/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowServiceTest.java
```diff
@@ -337,6 +337,22 @@ public void createCase_whenHasOrganisationCorrespondentDetails() {
...
+        UUID userUUID = UUID.randomUUID();
+        Map<String, String> receivedData = new HashMap<>();
+        MigrateCaseworkCaseResponse migrateCaseworkCaseResponse = new MigrateCaseworkCaseResponse(UUID.randomUUID(), null);
+
+        when(caseworkClient.migrateCase(any(), any(), any(), any())).thenReturn(migrateCaseworkCaseResponse);
+
+        MigrateCaseResponse output = workflowService.migrateCase(caseDataType, dateReceived, documents, userUUID,null, receivedData);
+        assertThat(output.getUuid()).isNotNull();
+        verify(camundaClient, times(1)).startCase(any(), any(), any());
+        verify(caseworkClient, times(0)).saveCorrespondent(any(), any(), any());
```
gary-newbolt, line 354: I don't think this test is really testing the logic of the method here... `migrateCase` does a whole bunch of data mapping logic that your test isn't catching. Neither is it verifying that the caseworkClient method to migrate is called.

src/test/java/uk/gov/digital/ho/hocs/workflow/api/dto/MigrateCaseResponseTest.java
```diff
@@ -0,0 +1,33 @@
...
+import org.junit.Test;
+
+import java.util.UUID;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class MigrateCaseResponseTest {
+
+    @Test
+    public void getMigrateCaseResponse() {
```
gary-newbolt, line 12: I'm not sure why you're bothering with this test, it looks like you're testing whether a java constructor works... we know they do, there's no extra logic in the constructor. I'd recommend removing it.

gary-newbolt (dismissed):
lgtm

gary-newbolt (approved):

PR #647, opened 2022-02-28
Hocs-4389: Remove unneeded healthcheck file

nickmcmahon01:
This file was being copied into the docker image, it is no longer referenced by the docker-compose so can be deleted.


LiamMacP (approved):
Change looks good - approved

PR #652, opened 2022-03-09
HOCS-4221: Add ability to transfer BF cases to SMC

JosephRogerson:


gary-newbolt (changes requested):
There's an errant gateway I think here...
Screenshot 2022-03-09 at 10 56 41

JosephRogerson:
> There's an errant gateway I think here... Screenshot 2022-03-09 at 10 56 41

I've removed that gateway

gary-newbolt:
Changes look good, just need to confirm which documents to transfer for BF --> SMC (PSU team)

gary-newbolt (approved):

PR #657, opened 2022-03-14
Feature/hocs 4718 to add realloc reason text area

abarker66:
Added a text area to add a reason for business area transfer.


gary-newbolt (approved):
lgtm

PR #659, opened 2022-03-15
HOCS-4761 add fix to allow case to move correctly from QA and Draft

abarker66:
Added a fix to QA and Draft stages to allow cases to transfer to Stop List and Campaign and reject to earlier stages.


gary-newbolt (approved):
lgtm

PR #670, opened 2022-03-18
HOCS-4632: support GRO triaging allocation

LiamMacP:
Changes to the POGR and the POGR_REGISTRATION flows to support the
movement of a case through to the triaging stages. For GRO cases a new
screen is shown that allows for the selection of a triaging team.


LiamMacP:
This needs the base changing to main before merging - #669 needs to be merged first.

robertdeniszczyc2 (approved):

PR #684, opened 2022-04-06
HOCS-4862: update match label action version 

RZX5678:
Update the match label action version to the latest release.


LiamMacP (approved):
Approved - change looks good!

PR #689, opened 2022-04-13
HOCS-4863: update semver tag action release version

RZX5678:
Update semver tag action version to the latest release.


LiamMacP (approved):
Approved - looks good!

PR #697, opened 2022-04-22
Remove db initialise initcontainer

nickmcmahon01:


LiamMacP (approved):
Approved - looks good!

PR #699, opened 2022-04-26
HOCS-3780: Remove diff based tests for workflows

nickmcmahon01:


build.gradle
```diff
@@ -12,25 +12,6 @@ repositories {
     mavenCentral()
 }
 
-plugins.withType(JacocoPlugin) {
```
nickmcmahon01, line 15: These are defined twice for some reason

gary-newbolt (approved):
Looks good

PR #703, opened 2022-05-09
HOCS-5000: stage duplication problem

gary-newbolt:


src/main/java/uk/gov/digital/ho/hocs/workflow/BpmnService.java
```diff
@@ -67,25 +67,17 @@ public String createStage(String caseUUIDString, String stageUUIDString, String
...
-
-            CreateCaseworkStageRequest stageRequest = new CreateCaseworkStageRequest(stageTypeString, teamUUID, userUUID, allocationType);
-            log.debug("Creating new stage {} for case {}", stageRequest, caseUUIDString);
-            resultStageUUID = caseworkClient.createStage(caseUUID, stageRequest).toString();
-            log.info("Created Stage {} for Case {}", resultStageUUID, caseUUID);
+            stageUUID= UUID.fromString(stageUUIDString);
         }
 
+        CreateCaseworkStageRequest stageRequest = new CreateCaseworkStageRequest(stageTypeString, stageUUID, teamUUID, userUUID, allocationType);
+        log.debug("Creating new stage {} for case {}", stageRequest, caseUUIDString);
```
LiamMacP, line 77: Are both this log and also log line 64 required? I sort of feel this one would be sufficient?

src/main/java/uk/gov/digital/ho/hocs/workflow/BpmnService.java
```diff
@@ -67,25 +67,17 @@ public String createStage(String caseUUIDString, String stageUUIDString, String
...
+        resultStageUUID = caseworkClient.createStage(caseUUID, stageRequest).toString();
+        log.info("Created Stage {} for Case {}", resultStageUUID, caseUUID);
```
LiamMacP, line 79: I know that this was probably the original log line - but could we pass through an event from LogEvent here. Helps with identification in OpenSearch

gary-newbolt: added

src/main/java/uk/gov/digital/ho/hocs/workflow/BpmnService.java
gary-newbolt: I've removed this line

LiamMacP (approved):

PR #717, opened 2022-05-25
HOCS-3976: remove case viewed audits

nickmcmahon01:
Hocs-3976: remove call causing case viewed audits when not needed

### DIFF
--- a/src/main/java/uk/gov/digital/ho/hocs/workflow/security/AllocatedAspect.java
+++ b/src/main/java/uk/gov/digital/ho/hocs/workflow/security/AllocatedAspect.java
@@ -9,6 +9,7 @@ import org.springframework.context.annotation.Conditional;
 import org.springframework.stereotype.Component;
 import uk.gov.digital.ho.hocs.workflow.application.NonMigrationEnvCondition;
 import uk.gov.digital.ho.hocs.workflow.client.caseworkclient.CaseworkClient;
+import uk.gov.digital.ho.hocs.workflow.client.infoclient.InfoClient;
 
 import java.util.Set;
 import java.util.UUID;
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/digital/ho/hocs/workflow/security/AllocatedAspect.java
```diff
@@ -82,9 +84,9 @@ private void verifyAllocatedToTeam(UUID caseUUID, UUID stageUUID) {
     }
 
     private boolean proceedIfUserTeamIsAdminForCaseType(UUID caseUUID) {
-        String stageType = caseworkClient.getCase(caseUUID).getType();
+        String shortCode = caseUUID.toString().substring(34);
```
LiamMacP, line 87: This still feels like a hack, can we add something to review this...

src/test/java/uk/gov/digital/ho/hocs/workflow/security/AllocatedAspectTest.java
```diff
@@ -25,6 +26,9 @@ public class AllocatedAspectTest {
     @Mock(answer = Answers.RETURNS_DEEP_STUBS)
     private CaseworkClient caseworkClient;
 
+    @Mock(answer = Answers.RETURNS_DEEP_STUBS)
```
LiamMacP, line 29: Does this need `(answer = Answers.RETURNS_DEEP_STUBS)`?

nickmcmahon01: 👻 whooo knowws, i'll take it out

src/main/java/uk/gov/digital/ho/hocs/workflow/security/AllocatedAspect.java
nickmcmahon01: there is an endpoint on casework we could add to workflow, but that endpoint calls the info service one. so we can unpick it as one task.

LiamMacP (approved):

PR #721, opened 2022-06-01
HOCS-5161: move away from custom 'name' endpoint

nickmcmahon01:
Move away from custom endpoint to help docs service handle database access in a consistent way


src/main/java/uk/gov/digital/ho/hocs/workflow/client/documentclient/dto/DocumentDto.java
```diff
@@ -0,0 +1,58 @@
...
+
+import java.time.LocalDateTime;
+import java.util.LinkedHashSet;
+import java.util.Set;
+import java.util.UUID;
+
+@AllArgsConstructor()
+@Getter
+@Slf4j
+public class DocumentDto {
```
LiamMacP, line 16: Do we need this entire DTO for the display name?

nickmcmahon01: no, I was hoping it would be useful as/when we move other calls over, but it can come in with those changes

PR #732, opened 2022-06-30
HOCS-5246 added reject note to draft stage

abarker66:
Addition of a draft stage rejection note.


LiamMacP:
Could you add a default flow from the exclusive gateway back to the screen for odd chance a screen doesn't respond with direction of `BACKWARDS` or `FORWARDS`? While its not necessary just future proofs it and makes sure if it fails its in a workable state. 

```
GRO: Gateway_0t2n8w2
HMPO: Gateway_0bt29iu
```

abarker66:
The DIRECTION value is defined after case creation and FORWARD is essentially the default value, the only time it is switched is when the back link is pressed - it is not a good idea to default the direction decision.

LiamMacP:
> The DIRECTION value is defined after case creation and FORWARD is essentially the default value, the only time it is switched is when the back link is pressed - it is not a good idea to default the direction decision.

That's not strictly true, we also use it for the flow-direction component. As I say not likely to be used, but we should be defensively coding.

LiamMacP:
> Yes it is used primarily for the flow-direction - it is not normal to default flow-direction decisions - I will add this (it is a bad idea) but you need to tell me which direction is the default and where does the default flow go back to????

It's just a default flow so doesn't need a value and back into the screen.

abarker66:
No I mean which flow (forward, backward) do you want as the default - because it will always be one or the other because the workflow would not get that far without the flow direction being set? - we will be defaulting a flow which will always have a value - it makes no sense

LiamMacP:
> No I mean which flow (forward, backward) do you want as the default - because it will always be one or the other because the workflow would not get that far without the flow direction being set? - we will be defaulting a flow which will always have a value - it makes no sense

I think this is where there is crossed wires, there should be a third one off the exclusive gateway that leads back to the screen. So forward and backwards stays the same.

abarker66:
Jesus H. Corbett - why did you not say that in the first place - you are aware it will never be called because of the way workflow works - it would never get to that decision point if direction was not not set - it is either forward or backward - no other tenant bpmn diagram defaults at this point - there are subsequent decisions that default but direction is dealt with first and not defaulted. I will add it so I can move on. If this is to be a new standard you have a lot of rework to do - about 160 bpmn diagrams will need updating - enjoy.

abarker66:
Added change and tested with Camunda test suite:
1) If Direction is not set at all then the Camunda engine fails followed by the Workflow back end - it needs Direction to exist.
2) If Direction exists but is not assigned a value as it leaves the activity then the default flow ensures it gets stuck in an infinite loop. It can only escape from once a value is assigned to Direction.
In light of what I'm seeing I think this suggestion is a bad idea given how workflow deals with flow-direction - I am going to return the ticket to the back log and let someone else have a go.

LiamMacP:
I have pushed a change to add the default flow back in. While we are safe to assume that DIRECTION will always be set to _something_ we cannot always gaurantee that this will be `BACKWARD` or `FORWARD`. Having the default route here will enable previous cases to still be workable without relying on the camunda engine to randomly pick which route to use (which could lead to undesirable behaviour). 

While this will mean potential manual intervention in the future - we should strive to ensure backwards compatability with cases.

abarker66:
Absolute nonsense - but knock yourself out.

robertdeniszczyc2 (approved):

nickmcmahon01:
Do we need to raise a ticket to ensure a default flow is present for all decision points?

robertdeniszczyc2:
> Do we need to raise a ticket to ensure a default flow is present for all decision points?

It's worth a huddle; can we do something in the code which executes it to avoid having to put in default flows maybe? Thinking in the updateStage method in WorkflowService

nickmcmahon01:
> > Do we need to raise a ticket to ensure a default flow is present for all decision points?
> 
> It's worth a huddle; can we do something in the code which executes it to avoid having to put in default flows maybe? Thinking in the updateStage method in WorkflowService

I don't think so, its not a fun task for anyone though is it - but who knows what else we'll find!

PR #735, opened 2022-07-07
HOCS-5323: add back behaviour for GRO Draft

LiamMacP:
Add missing back button handling for GRO draft cases. There is also an
addition of a couple of missing tests that let this slip through the net
last time.


src/test/java/uk/gov/digital/ho/hocs/workflow/processes/pogr/POGR_GRO_DRAFT.java
```diff
@@ -87,11 +87,13 @@ public void testTelephoneResponse() {
     @Test
     public void testNotTelephoneResponse() {
         whenAtCallActivity("POGR_GRO_PRIORITY_CHANGE_SCREEN")
+                .thenReturn("DraftOutcome", "TelephoneResponse")
```
vinothon84, line 90: Hi @LiamMacP - I see line 90 and 91 are duplicated. Is this expected?

LiamMacP: Hi @vinothon84 - yes you have to specify what should happen on each trip into the process `POGR_GRO_PRIORITY_CHANGE_SCREEN`. It allows you to change routes if you require - in our case we don't need to.

vinothon84 (approved):

PR #744, opened 2022-07-27
Hocs 4943 ability to close case at draft stage

vinothon84:
Hocs 4943 - ability to close case at draft stage
1) Covers both scenarios - POGR HMPO and GRO (Complaints & Policy)
2) BPMN file changes to add POGR_HMPO_Draft & POGR_GRO_Draft - Close the case - BPMN file changes
3) BPMN file changes to cover flow when it passed it back to POGR_HMPO & POGR_GRO
4) Junit test suits to covers new flow for Draft close case for HMPO & GRO

PR #745, opened 2022-07-28
HOCS-4943: support draft case closure 

vinothon84:
HOCS-4943 for POGR - Closing the case @ Draft State

1. DML changes to add Close the case radio button
2. Add Close Reason drop down values
3. Creation of new screen - with label, inset, text area, Complete case and back button
4. Respective test cases were created.

Opening this PR after closing - https://github.com/UKHomeOffice/hocs-workflow/pull/744


JosephRogerson (changes requested):
This might be a bit better if in POGR_HMPO and POGR_GRO the flow connected to the final complete case

Also, could you try and see if it still works without the complete case action in the _DRAFT files? You may not need this any more now that the flow is connected correctly
image

JosephRogerson (approved):

PR #749, opened 2022-08-04
HOCS-5428: Fix POGR2 escalation

JosephRogerson:


abarker66 (approved):
LGTM, tested and I can now see the new stage 2 cases at data input stage.

PR #750, opened 2022-08-05
HOCS-5363 add global form and summary

LiamMacP:
Add the ability to add a summary to the form schema and have it
propagate through the service.

As `SchemaDtoBuilder` file is currently only used as a helper within
the test files, this change moves this to a helper folder within the
test directory. This change also removes the static-ness of this file as
not required.

Add endpoint and necessary service to support the retrieving of a form
that is outside a stage (effectively global).


src/main/java/uk/gov/digital/ho/hocs/workflow/api/FormResource.java
```diff
@@ -0,0 +1,31 @@
...
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RestController;
+import uk.gov.digital.ho.hocs.workflow.api.dto.GetFormResponse;
+
+import java.util.UUID;
+
+import static org.springframework.http.MediaType.APPLICATION_JSON_VALUE;
+
+@RestController
+@Slf4j
```
LiamMacP, line 16: This isn't required.

src/main/java/uk/gov/digital/ho/hocs/workflow/api/FormService.java
```diff
@@ -0,0 +1,44 @@
...
+import uk.gov.digital.ho.hocs.workflow.domain.model.forms.HocsFormAccordion;
+import uk.gov.digital.ho.hocs.workflow.domain.model.forms.HocsFormField;
+import uk.gov.digital.ho.hocs.workflow.domain.model.forms.HocsFormSecondaryAction;
+import uk.gov.digital.ho.hocs.workflow.domain.model.forms.HocsSchema;
+
+import java.util.List;
+
+import java.util.UUID;
+
+@Service("HocsFormService")
```
LiamMacP, line 19: Camunda registers a bean of FormService, this allows us to keep the sane name but just flag this in Spring land with this name.

src/main/java/uk/gov/digital/ho/hocs/workflow/api/FormResource.java
robertdeniszczyc2:
```suggestion
```

LiamMacP: Why didn't I think of that - I was putting off pushing a change 😆

src/main/java/uk/gov/digital/ho/hocs/workflow/api/FormResource.java
```diff
@@ -0,0 +1,31 @@
+package uk.gov.digital.ho.hocs.workflow.api;
+
+import lombok.extern.slf4j.Slf4j;
```
LiamMacP, line 3:
```suggestion

```

robertdeniszczyc2 (approved):

PR #760, opened 2022-08-17
HOCS-5471: rework investigation diagrams

LiamMacP:
Rework the POGR GRO and HMPO diagrams from using the word Triage to the
more appropriate Investigation. This also adds default flows for all
screens and necessary variable checks to ensure we can progress. A new
back flow has also been added between the Investigation screen and
Accept Case screens.


LiamMacP:
This is a breaking change for POGR but as they aren't live this is fine.

JosephRogerson (approved):

PR #762, opened 2022-08-18
HOCS-5469: rework case closure

LiamMacP:
Rework of the case closure mechanic for POGR workflow. This change
includes pulling the case closure into a generic process that is called
by HMPO and GRO Draft/Investigation screens.


LiamMacP:
Marking as minor as while it breaks POGR cases they aren't live and flows are subject to change.

JosephRogerson (approved):

PR #769, opened 2022-08-22
HOCS-5031: POGR2 interim letters

JosephRogerson:


LiamMacP (approved):
I think we decided to cull the [Stage1]_[Stage2].diff tests. Fancy removing them all and if there is still a presence within the build.gradle, bin that too 🙂

LiamMacP:
If you could either squash or cleanup the history on merge that would be appreciated too. If you need a hand let me know...

JosephRogerson:
> 

These are already gone. Not sure how I accidentally re-added one here, must have had it in my git stash or something

PR #790, opened 2022-09-14
HOCS-5512: IEDET capture additional Complaint Details

AlexW013:


src/main/resources/processes/IEDET/IEDET_TRIAGE.bpmn
```diff
@@ -1,12 +1,13 @@
 <?xml version="1.0" encoding="UTF-8"?>
```
robertdeniszczyc2, line 1: Does the Serious Complaint Details screen need a BACKWARD route to Complaint Category?

Also would name the tasks Serious/<not serious, unsure of the proper name> Complaint Details for easier readability rather than both being Complaint Details

robertdeniszczyc2 (commented):
Added a question and a small suggestion to IEDET_TRIAGE

Cheers,
Rob

src/main/resources/processes/IEDET/IEDET_TRIAGE.bpmn
AlexW013: Regarding the BACKWARD route, I get what you mean. I was going off the pattern from the earlier screens where the path seems to go from the gateway back to the screen itself but that didn't make much sense to me. So should they all in theory have a route back to the previous screen?

Screenshot 2022-09-16 at 10 01 57

robertdeniszczyc2: It depends on whether this screen has a back link in the UI to take the user back to the last screen. If there's a back link then we should handle that in the workflow, if one isn't included then we don't need the backward path.

AlexW013: Okay. The sample screenshots on the ticket do include a back link so we probably should add one

robertdeniszczyc2: Previously `Flow_0kbmn7q` went to `Screen_ComplaintDetails_NotSerious` but now it goes to `Screen_ComplaintCategory`, which `Flow_10fh17c` also follows the same path, not sure if this is desired as there are two flows with the same behaviour now?

AlexW013: Okay yes, I get it now. Refactored so they each have a green flow with the BACKWARDS condition and a red flow with no condition

PR #801, opened 2022-10-01
HOCS-5281: migrate field definitions to service

LiamMacP:
Refactor of the screens used within the system to bring them into the service. This effectively removes the complex SQL changes currently required to propagate a screen change. This simplifies the system by allowing easy changes to be handled within JSON files and patched as necessary.

To aid this a JsonFileReader has been added, this at runtime allows for screens to be loaded by the system and using the current infrastructure return the same response payload as before (supporting backwards compatibility).


src/main/resources/screens/COMP_HIDDEN.json
```diff
@@ -0,0 +1,37 @@
+{
```
LiamMacP, line 1: Screens like these aren't actually required anymore as summary, accordion and extracts all have their own config. Removing this aren't in scope of this change though - as we need to ensure each isn't used _somewhere_...

src/main/java/uk/gov/digital/ho/hocs/workflow/domain/repositories/JsonFileReader.java
```diff
@@ -0,0 +1,35 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.springframework.stereotype.Service;
+import uk.gov.digital.ho.hocs.workflow.application.LogEvent;
+import uk.gov.digital.ho.hocs.workflow.domain.exception.ApplicationExceptions;
+
+import java.io.IOException;
+import java.io.InputStream;
+
+@Service
+public abstract class JsonFileReader {
```
LiamMacP, line 13: I'm not actually convinced this is a useful helper - we could and probably should role this into the ScreenRepository

src/main/java/uk/gov/digital/ho/hocs/workflow/domain/repositories/JsonFileReader.java
```diff
@@ -0,0 +1,35 @@
...
+        this.objectMapper = objectMapper;
+    }
+
+    protected <T> T readValueFromFile(TypeReference<T> reference, String fileName) {
+        try (InputStream in = Thread.currentThread().getContextClassLoader().getResourceAsStream(
+            String.format("%s/%s.json", getFolderName(), fileName))) {
+            return objectMapper.readValue(in, reference);
+        } catch (IOException e) {
+            e.printStackTrace();
+            throw new ApplicationExceptions.JsonFileReadException(
```
LiamMacP, line 27: This also supports moving into the ScreenRepository - we should have a proper Exception...

LiamMacP (commented):
This is a very _dumb_ implementation - we could be cleverer with directories for the screens - merging with respective camunda diagrams. We would have to take care with shared screens - but symbolic links should help there. Out of scope for this though.

src/main/java/uk/gov/digital/ho/hocs/workflow/domain/model/forms/HocsFormField.java
```diff
@@ -33,13 +35,26 @@ public static HocsFormField from(FieldDto fieldDto) {
         return new HocsFormField(fieldDto.getComponent(), fieldDto.getValidation(), props);
     }
 
+    public static HocsFormField from(Field field) {
```
LiamMacP, line 38: This is a dupe of the above with just a different method parameter - we could be smarter but as it is deprecated I don't see the point.

src/main/java/uk/gov/digital/ho/hocs/workflow/domain/repositories/ScreenRepository.java
```diff
@@ -0,0 +1,25 @@
...
+import uk.gov.digital.ho.hocs.workflow.domain.repositories.entity.Schema;
+
+@Service
+public class ScreenRepository extends JsonFileReader {
+
+    protected ScreenRepository(ObjectMapper objectMapper) {
+        super(objectMapper);
+    }
+
+    //TODO: investigate cachable benefits
```
LiamMacP, line 15: Having run this locally some of the files took around 0.02ms to load, with repeat loads of < 0.005ms. I'm not convinced caching is necessary for this at the moment.

SWhitlam (approved):

PR #803, opened 2022-10-06
HOCS-5684: Add get case for team endpoint

JosephRogerson:


src/main/resources/screens/DCU_OFFLINE_QA_DETAILS.json
```diff
@@ -8,7 +8,7 @@
         "required"
       ],
       "props": {
-        "choices": "USERS_FOR_CASE"
+        "choices": "USERS_FOR_STAGE"
```
LiamMacP, line 11: It would be important for it to be noted that this change will have to go after the hocs-frontend/ hocs-casework ones due to the functionality required.

LiamMacP: Maybe add it on the ticket?

LiamMacP (approved):

src/main/resources/screens/DCU_OFFLINE_QA_DETAILS.json
JosephRogerson: Added as a comment on the ticket

PR #814, opened 2022-10-17
HOCS-5652 - json changes to update HMPO business areas

vinothon84:
HOCS-5652 : This PR contains json changes to load both POGR HMPO and GRO business areas. Currently DECS have common business areas which is same across both HMPO and GRO case types. As part of this change, we would have separate business areas for POGR/POGR2 HMPO and GRO case complaints.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -15,7 +15,7 @@ repositories {
 
 dependencyManagement {
     imports {
-        mavenBom 'org.camunda.bpm:camunda-bom:7.17.0'
+        mavenBom 'org.camunda.bpm:camunda-bom:7.18.0'
     }
 }
 
@@ -34,7 +34,7 @@ dependencies {
 
     implementation('net.logstash.logback:logstash-logback-encoder:7.2')
 
-    implementation('org.camunda.bpm:camunda-engine-spring:7.17.0')
+    implementation('org.camunda.bpm:camunda-engine-spring:7.18.0')
     implementation('com.fasterxml.uuid:java-uuid-generator:4.0.1')
     implementation('commons-codec:commons-codec:1.15')
     implementation('org.apache.httpcomponents:httpclient:4.5.13')


src/main/resources/screens/POGR_BUS_CONTRIBUTIONS_ADD.json
```diff
@@ -8,12 +8,35 @@
         "required"
       ],
       "props": {
-        "choices": "POGR_CONTRIB_BUS_AREA"
+        "choices": "POGR_CONTRIB_BUS_AREA",
```
LiamMacP, line 11: Would this be better using the conditionalChoices mechanic here than introducing a whole new field?

LiamMacP: Especially as this is the only thing changing?

vinothon84: Apart from this - name and label shall be included as well. Do we really need to using condition to render all 3 properties (choices, name and label) ?

LiamMacP: No you just need to change the choices that are shown to the user.

vinothon84: I have incorporated your comment and it makes code tidy as well 👍

src/main/resources/screens/POGR_BUS_CONTRIBUTIONS_ADD.json
```diff
@@ -1,41 +1,27 @@
 {
-  "props": {},
-  "title": "Add Contribution Request",
```
LiamMacP, line 3: Why have these been removed?

LiamMacP (approved):

PR #834, opened 2022-11-09
[HOCS-5883] Correct typo in PSU triage stage CloseCase option

jeff-horton-ho-sas:
In the IEDET PSU Triage stage, the option 'No - send to team not on DECS' is instead showing as 'No - send to team not to DECS'


LiamMacP (approved):
Changes look good - it would be worth putting the ticket number within the commit message so in the history we can easily see what instigated this change.

LiamMacP:
You will also need to add a label to the PR - `patch`, `minor`, or `major` based on the SemVer rating of the change. This would fit into a `patch`.

PR #835, opened 2022-11-10
HOCS-5872: support additional screen folders

LiamMacP:
Support the ability to load screens from the multiple folders based on an inputted list. This first checks the folders in order of specificity and then defaults to the live folders.

This also includes a minor refactoring to remove some code duplicity.


src/main/java/uk/gov/digital/ho/hocs/workflow/api/FormService.java
```diff
@@ -22,25 +26,41 @@ public class FormService {
...
+    private final List<String> additionalScreenFolders;
+
+    public FormService(CaseworkClient caseworkClient,
+                       ScreenRepository screenRepository,
+                       @Value("${hocs.screens.additionalFolders}") List<String> fieldFolders) {
         this.caseworkClient = caseworkClient;
         this.screenRepository = screenRepository;
+
+        this.additionalScreenFolders = fieldFolders;
+        this.additionalScreenFolders.add("live"); //TODO: magic word - can/do we extract this?
```
LiamMacP, line 38: Conversation to be had about whether we just have this as a default here - or we embed it directly into the deployment properties.

SWhitlam (approved):

PR #843, opened 2022-11-17
Hocs 5851 iede tto psu amend sla

vinothon84:
HOCS-5851 - This PR would help to update case 
deadline to 84 days for PSU stages which will
be displayed in Summary page whereas if its
sent it back to IEDET then its set to 20 days. 
Currently this deadline date is unaltered and will 
be created during case creation irrespective of 
case pushed to PSU or send it back to IEDET.


src/main/resources/processes/IEDET/IEDET_PSU_TRIAGE.bpmn
```diff
@@ -29,20 +29,25 @@
...
       </bpmn:extensionElements>
-      <bpmn:incoming>Flow_02ks0ls</bpmn:incoming>
+      <bpmn:incoming>Flow_18kgsnq</bpmn:incoming>
       <bpmn:outgoing>Flow_09pdaeg</bpmn:outgoing>
     </bpmn:task>
     <bpmn:sequenceFlow id="Flow_09pdaeg" sourceRef="Activity_1u4jmf5" targetRef="EndEvent_Triage" />
     <bpmn:sequenceFlow id="Flow_1gidqty" name="" sourceRef="Gateway_0khr7oc" targetRef="EndEvent_Triage">
       <bpmn:conditionExpression xsi:type="bpmn:tFormalExpression">${PsuTriageOutcome == "CloseCase"}</bpmn:conditionExpression>
     </bpmn:sequenceFlow>
+    <bpmn:serviceTask id="Service_UpdateDeadline" name="Update Case Deadline" camunda:expression="${bpmnService.updateDeadlineDays(execution.getVariable(&#34;CaseUUID&#34;),execution.processBusinessKey,20)}">
```
JosephRogerson, line 46: I think it might be a bit better if this action was in PSU_IEDET_COMPLAINT instead then it wouldn't have to be duplicated for triage and outcome

vinothon84: Incorporated the changes

src/main/resources/processes/IEDET/IEDET_PSU_OUTCOME.bpmn
```diff
@@ -154,6 +157,15 @@
       <bpmndi:BPMNShape id="Activity_1an6g5s_di" bpmnElement="Activity_ResetCategories">
         <dc:Bounds x="460" y="200" width="100" height="80" />
       </bpmndi:BPMNShape>
+      <bpmndi:BPMNShape id="Event_0iu7swi_di" bpmnElement="EndEvent_Outcome">
+        <dc:Bounds x="832" y="319" width="36" height="36" />
+      </bpmndi:BPMNShape>
+      <bpmndi:BPMNShape id="Activity_15bvtsg_di" bpmnElement="Activity_ReturnCase">
+        <dc:Bounds x="720" y="200" width="100" height="80" />
+      </bpmndi:BPMNShape>
+      <bpmndi:BPMNShape id="Activity_1f3oa3j_di" bpmnElement="Service_UpdateDeadline">
```
JosephRogerson, line 166: This can also be moved to PSU_IEDET_COMPLAINT, just after the PSU activity so that it looks like this:
image
That way you will only need to change that file and no others

JosephRogerson (approved):

PR #845, opened 2022-11-22
[HOCS-5694] Remove gender radios from complainant details

jeff-horton-ho-sas:
https://collaboration.homeoffice.gov.uk/jira/browse/HOCS-5694


LiamMacP (commented):
Does this need removing from [audit](https://github.com/UKHomeOffice/hocs-audit/blob/main/src/main/resources/config/case-data-fields.json) extracts, [unallocated views](https://github.com/UKHomeOffice/hocs-casework/tree/main/src/main/resources/config/details/stages), and [summary](https://github.com/UKHomeOffice/hocs-casework/tree/main/src/main/resources/config/summary)?

LiamMacP (approved):

PR #846, opened 2022-11-23
HOCS-5538: IEDET PSU review categories

JosephRogerson:


JosephRogerson:
> Would these screens be better with the IEDET prefix - or is there a good presumption that these are a default PSU screen?

This screen appears in the UKVI PSU flow and has the same categories so I think it is fair to keep this as default for now with the idea that it may have to change if the screens diverge

LiamMacP (approved):

PR #856, opened 2022-11-30
HOCS-5564: Implement business events

JosephRogerson:


src/main/java/uk/gov/digital/ho/hocs/workflow/api/bpmn/BusinessEventService.java
```diff
@@ -0,0 +1,98 @@
...
+        String name = String.format(DATA_FIELD_UPDATED_EVENT_NAME, fieldName);
+
+        GetCaseworkCaseDataResponse caseData = caseworkClient.getCase(caseUUID);
+        String previousValue = caseData.getData().get(previousValueKey);
+
+        if (Objects.equals(previousValue, newValue)) {
+            return;
+        }
+
+        GetStageResponse getStageResponse = workflowService.getStage(caseUUID, stageUUID);
```
SWhitlam, line 56: I don't think we should call getStage here. It gets the case from casework which duplicates line 49 with an additional call to casework and it also calls casework to assign the user as a side effect of getting the stage (which is a code smell on it's own).
As an alternative consider calling the FormService to get the schema. Either GetFormSchema with the form name or consider refactoring to use getFormWithCase which would allow you to get the case data at the same time so you could replace line 49 too.

src/main/java/uk/gov/digital/ho/hocs/workflow/api/bpmn/BusinessEventService.java
```diff
@@ -0,0 +1,98 @@
...
+                                CaseworkClient caseworkClient,
+                                BpmnService bpmnService,
+                                WorkflowService workflowService) {
+        this.auditClient = auditClient;
+        this.caseworkClient = caseworkClient;
+        this.bpmnService = bpmnService;
+        this.workflowService = workflowService;
+    }
+
+    public void createDataFieldUpdatedEvent(String caseUUIDString, String stageUUIDString, String fieldName, String newValue) {
```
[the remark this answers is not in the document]
jeff-horton-ho-sas: We tried this at the start, and there wasn't an obvious way to get the label data into the front end without an additional request to get the value => label mapping for each event. Happy to discuss if you have an idea for how to get that data.

src/test/java/uk/gov/digital/ho/hocs/workflow/BpmnServiceTest.java
```diff
@@ -887,4 +891,75 @@ public void testShouldThrowIfVarArgsAreOddNumber() {
 
     }
 
+ /*   @Test
```
SWhitlam, line 894: do these need removing or updating?

src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowConstants.java
```diff
@@ -29,5 +29,6 @@ public interface WorkflowConstants {
     String CHANNEL = "Channel";
     String CHANNEL_COMP_ORIGINATEDFROM = "XOriginatedFrom";
     String CHANNEL_COMP_WEBFORM = "Webform";
-
+    String PREVIOUS_VALUE_KEY = "previous%s";
```
LiamMacP, line 32:
```suggestion
    String PREVIOUS_VALUE_KEY = "XPrevious%s";
```

This would fit the `X...` pattern for values that should not be used outside of workflow.

src/main/java/uk/gov/digital/ho/hocs/workflow/api/WorkflowConstants.java
```diff
@@ -29,5 +29,6 @@ public interface WorkflowConstants {
...
+    String DATA_FIELD_UPDATED_EVENT_NAME = "%s_UPDATED";
```
LiamMacP, line 33:
```suggestion
    String DATA_FIELD_UPDATED_EVENT_NAME = "CASE_DATA_ITEM_UPDATED";
```

Would this not just be this - we would have enough in the FE then to string interpolation for the UI.

src/main/java/uk/gov/digital/ho/hocs/workflow/api/bpmn/BusinessEventService.java
```diff
@@ -0,0 +1,98 @@
...
+    private String getLabel(String value, List<Map<String, String>> choices) {
+        return choices.stream()
+            .filter(choice -> Objects.equals(choice.get("value"), value))
+            .findFirst()
+            .map(choice -> choice.get("label"))
+            .orElse(value);
+    }
+
+    private <T> T getFormProperty(String name, Optional<HocsFormField> hocsFormField, T defaultValue, Class<T> classTag) {
+        return hocsFormField
```
SWhitlam, line 92: This needs to take conditional values into account. You need to check for conditional field props and use that if there is one, otherwise default to the non conditional field.

src/main/java/uk/gov/digital/ho/hocs/workflow/client/auditclient/AuditClient.java
```diff
@@ -0,0 +1,113 @@
...
+    private final String raisingService;
+    private final String namespace;
+    private final AmazonSNSAsync auditSnsClient;
+    private final ObjectMapper objectMapper;
+    private final RequestData requestData;
+    private final static String EMPTY_JSON = "{}";
+
+    @Autowired
+    public AuditClient(AmazonSNSAsync auditSnsClient,
+                       @Value("arn:aws:sns:eu-west-2:000000000000:hocs-audit-topic") String auditQueue,
```
LiamMacP, line 42: This is hardcoded? this should be pulled from the application properties?

src/main/java/uk/gov/digital/ho/hocs/workflow/client/auditclient/AuditClient.java
```diff
@@ -0,0 +1,113 @@
...
+                       @Value("wf:hocs-workflow-service") String raisingService,
```
LiamMacP, line 43: Same for this - should be in the application properties

src/main/java/uk/gov/digital/ho/hocs/workflow/application/aws/SnsConfiguration.java
```diff
@@ -0,0 +1,28 @@
...
+import com.amazonaws.services.sns.AmazonSNSAsync;
+import com.amazonaws.services.sns.AmazonSNSAsyncClientBuilder;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Primary;
+import org.springframework.context.annotation.Profile;
+
+@Configuration
+@Profile({ "sns" })
```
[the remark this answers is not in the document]
JosephRogerson: PR to add to helm charts is here:
https://github.com/UKHomeOffice/hocs-helm-charts/pull/53

JosephRogerson:
@SWhitlam/ @LiamMacP, we've resolved the comments here, please could you have another look when you get the chance

src/main/java/uk/gov/digital/ho/hocs/workflow/api/bpmn/BusinessEventService.java
```diff
@@ -0,0 +1,112 @@
...
+            newValueLabel,
+            previousValueLabel
+        );
+
+        auditClient.createBusinessEvent(caseUUID, stageUUID, dataFieldUpdatedPayload);
+        bpmnService.updateCaseValue(caseUUIDString, stageUUIDString, previousValueKey, newValue);
+    }
+
+    private List<Map<String, String>> getChoices(Optional<HocsFormField> hocsFormField, GetCaseworkCaseDataResponse caseData) {
+        @SuppressWarnings("unchecked") List<Object> conditionChoices = (List<Object>) getFormProperty("conditionChoices", hocsFormField, List.class).orElse(List.of());
```
LiamMacP, line 83: Not blocking but this feels like a semi-odd way of doing this casting. We have places that we use `TypeReference<>`, so wonder if that can be used to simplify this.

LiamMacP (approved):

SWhitlam (approved):

src/main/java/uk/gov/digital/ho/hocs/workflow/api/bpmn/BusinessEventService.java
jeff-horton-ho-sas: https://github.com/UKHomeOffice/hocs-workflow/pull/856/commits/bcc5db725dcc52471e3e0ae8fe23a40deae59578

Migrated to use TypeReferences / objectMapper.

PR #857, opened 2022-12-01
Hocs 5566 rejected by ps utag

vinothon84:
HOCS-5566 - New tagging service is created which is 
similar to BPMNService. This tagging service would
be call hocs-casework tag end point in order to 
create a tag for the case. BPMN workflow changes 
been made for IEDET workflow and in future this 
would be extended for other workflows.

### DIFF
--- a/src/main/java/uk/gov/digital/ho/hocs/workflow/client/caseworkclient/CaseworkClient.java
+++ b/src/main/java/uk/gov/digital/ho/hocs/workflow/client/caseworkclient/CaseworkClient.java
@@ -283,7 +283,8 @@ public class CaseworkClient {
         log.info("Data for caseUUID: {} mapped successfully.", caseUUID);
     }
 
-    public void createCaseTag(UUID caseUUID, CaseTagRequest caseTagRequest) {
+    public void createCaseTag(UUID caseUUID, String tag) {
+        CaseTagRequest caseTagRequest = new CaseTagRequest(tag);
         restHelper.post(serviceBaseURL, String.format("/case/%s/tag", caseUUID), caseTagRequest, String.class);
     }
 }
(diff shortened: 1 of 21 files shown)


src/main/resources/processes/PSU/PSU_IEDET_COMPLAINT.bpmn
```diff
@@ -46,6 +46,11 @@
       <bpmn:outgoing>Flow_10v7ld8</bpmn:outgoing>
     </bpmn:exclusiveGateway>
     <bpmn:sequenceFlow id="Flow_0mvcbzc" sourceRef="Gateway_0n7b72e" targetRef="EndEvent_Complaint" />
+    <bpmn:serviceTask id="Service_AddCaseTag" name="Add a case tag" camunda:expression="${taggingService.createTagForCase(execution.getVariable(&#34;CaseUUID&#34;), &#34;REJECTED BY PSU&#34;)}">
```
LiamMacP, line 49: Could we change this to `REJECTED_BY_PSU` and add the mapping change on the FE to this ticket. It would also allow for us to put custom colours in as well and hit that AC.

src/main/java/uk/gov/digital/ho/hocs/workflow/client/caseworkclient/CaseworkClient.java
```diff
@@ -281,4 +283,8 @@ public void mapCaseData(UUID caseUUID, Map<String, String> keyMappings) {
         log.info("Data for caseUUID: {} mapped successfully.", caseUUID);
     }
 
+    public void createCaseTag(UUID caseUUID, String tag) {
+        CaseTagRequest caseTagRequest = new CaseTagRequest(tag);
+        restHelper.post(serviceBaseURL, String.format("/case/%s/tag", caseUUID), caseTagRequest, String.class);
```
LiamMacP, line 288:
```suggestion
        restHelper.post(serviceBaseURL, String.format("/case/%s/tag", caseUUID), caseTagRequest, Void.class);
```

We're not utilising the return here so I don't think we want to try and convert it...

LiamMacP (changes requested):
Just one small alteration - good work though!

src/main/java/uk/gov/digital/ho/hocs/workflow/client/caseworkclient/CaseworkClient.java
vinothon84: Incorporated changes.

PR #861, opened 2022-12-09
HOCS-5667: COMP add SMC radio option

JosephRogerson:


JosephRogerson:
> For me the screen should be within a beta screen folder. I'm not overly picky about that and maybe this makes it clear what the folder is for.
> 
> As this should only be used within local environments, this should follow the convention of being in a `application-localstack.yml` file?
> 
> To enable this for testing you'll want to add it to the QA Workflow helm deployment: https://github.com/UKHomeOffice/hocs-helmfile/blob/main/environments/qa/values/hocs-workflow.yaml.gotmpl. There are many ways this could be done though, reach out if you want to chat about it 🙂

We chose to have a separate folder because it allows us to keep the different workstreams separate

If we use localstack as the configuration then all developers will have these changes appear when developing locally which they may not want.

We're not sure what the release plan is for this but it seems like it may be beneficial to be able to turn this on in test and live by enabling a profile, rather than having to copy all the new screens into live to release

Happy to have a call to discuss this tomorrow

LiamMacP (approved):

PR #868, opened 2022-12-16
HOCS-5728: Add PSU registration to COMP

JosephRogerson:


src/main/resources/processes/COMP/COMP.bpmn
```diff
@@ -657,449 +660,499 @@
       <bpmn:outgoing>Flow_05362gy</bpmn:outgoing>
     </bpmn:serviceTask>
     <bpmn:sequenceFlow id="Flow_05362gy" sourceRef="Service_UpdateCCHDeadline" targetRef="CallActivity_COMP_RETURNS" />
+    <bpmn:sequenceFlow id="Flow_0cpwrpp" sourceRef="ExclusiveGateway_11h64pd" targetRef="Activity_0jvlgp0">
```
LiamMacP, line 663: Is it worth changing this to have a useful ID? `CallActivity_COMP_PSU` maybe. This will also need doing for others too.

LiamMacP (approved):
Approved with potential remedial comment - good work.

PR #872, opened 2022-12-19
HOCS-5677: COMP PSU add complaint outcome screen

JosephRogerson:


LiamMacP (commented):
Is `Flow_1l3jiab` and `Flow_14rll2q` useful here? Is there a further ticket to do anything more from this that I am missing? 

This probably would be useful to have tests on `COMP2_PSU_OUTCOME` too?

JosephRogerson:
> Is `Flow_1l3jiab` and `Flow_14rll2q` useful here? Is there a further ticket to do anything more from this that I am missing?
> 
> This probably would be useful to have tests on `COMP2_PSU_OUTCOME` too?

The flows are used when the final response screen is added

I think COMP2 tests would be valuable, there aren't any at the moment. It may be worth a ticket to add these?

JosephRogerson:
merged in https://github.com/UKHomeOffice/hocs-workflow/pull/884

PR #894, opened 2023-01-09
HOCS-5705 - Add recategorise workflow - complaint type screen

jeff-horton-ho-sas:
https://collaboration.homeoffice.gov.uk/jira/browse/HOCS-5705


jeff-horton-ho-sas:
Merged by #895

PR #900, opened 2023-01-18
HOCS-5732 - Amend SLA for UKVI-PSU & vice-versa

vinothon84:
HOCS-5732 - Currently SLA by default is set to 
20 working days. This PR would help to update SLA 
to 60 working days when case is accepted by PSU and 
when its moved to PSU outcome stage. When the user
reject this case and send it back to UKVI - then
SLA would be updated to 20 working days. Changes
are covered using BPMN changes similar to IEDET 
to PSU workflow stages.


vinothon84:
Closing PR - as this is address in another open PR - https://github.com/UKHomeOffice/hocs-workflow/pull/901
Syncing with master branch.

PR #919, opened 2023-01-31
HOCS-4538: Reason for closing field required for COMP triage input screens

AlexW013:


JosephRogerson (commented):
Does this need changes to the screens in the ukvi-psu directory? I'm not sure if these screens are in there, but if they are the change will also need making there

AlexW013:
> Does this need changes to the screens in the ukvi-psu directory? I'm not sure if these screens are in there, but if they are the change will also need making there

You're right, yes they are so I have repeated the changes in the ukvi-psu folder

PR #920, opened 2023-01-31
Hocs 5910 bf psu sla

vinothon84:
HOCS-5910 - This PR would increment deadline to 60 working days
when case will be escalated from BF to PSU outcome stage. When
case sent back to BF - again deadline is set back to 12 working days.
This PR include other changes for BF-PSU escalations 
(feature/hocs-5916-bf-psu-complaint-categories-for-recategorise) and 
I discussed with @JosephRogerson to confirm this.


JosephRogerson:
Merged in #925

PR #924, opened 2023-02-01
Hocs 5929 remove complaint type category

vinothon84:
HOCS-5929-This PR will include changes to reset 
complaint type, category and add REJECTED_BY_PSU 
tag when the case is sent from PSU to BF queue.


JosephRogerson:
Merged in #925

PR #926, opened 2023-02-03
Enable flags on dev environment

JosephRogerson:


LiamMacP (approved):
Tested locally - appears to be picked up correctly.

PR #930, opened 2023-02-06
HOCS-6214: Fix failing recategorisations from PSU

JosephRogerson:
This fixes the failing recategorisation which is caused by the `Stage` variable being empty


jeff-horton-ho-sas (changes requested):
Is it worth adding verify checks that the new stages are visited in the COMP_RECATEGORISE and BF_RECATEGORISE tests? This will catch if future changes acidentally cause the set stage or update team to be missed.

vinothon84 (approved):

jeff-horton-ho-sas (approved):
👍

PR #939, opened 2023-02-13
HOCS-6189:UI improvements for record case response

vinothon84:
This PR changes would include UI improvements for record case response
screen.

### DIFF
--- a/src/main/resources/screens/live/IEDET_SEND_INPUT.json
+++ b/src/main/resources/screens/live/IEDET_SEND_INPUT.json
@@ -50,7 +50,7 @@
           },
           {
             "label": "Other",
-            "value": "other"
+            "value": "Other"
           }
         ],
         "className": "govuk-radios--small"
@@ -67,7 +67,7 @@
         "visibilityConditions": [
           {
             "conditionPropertyName": "CaseOutcome",
-            "conditionPropertyValue": "other"
+            "conditionPropertyValue": "Other"
           }
         ]
       },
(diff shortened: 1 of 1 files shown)


src/main/resources/screens/live/IEDET_SEND_INPUT.json
```diff
@@ -22,18 +24,40 @@
...
-      "label": "Upload the Final Response"
+      "label": "Upload the final response"
     },
     {
       "validation": [
         "required"
       ],
       "props": {
-        "choices": "S_IEDET_CASE_OUTCOME"
+        "choices": [
```
JosephRogerson, line 34: I think this value also appears in the accordion and summary, so these choices will also need updating in the configs for those in casework.

src/main/resources/screens/live/IEDET_SEND_INPUT.json
```diff
@@ -22,18 +24,40 @@
...
+            "label": "Substantiated",
+            "value": "substantiated"
+          },
+          {
+            "label": "Unsubstantiated",
+            "value": "unsubstantiated"
+          },
+          {
+            "label": "Partially Substantiated",
+            "value": "partiallysubstantiated"
```
JosephRogerson, line 45: Would having the values in pascal case be a bit more consistent with the dropdowns in other places?

src/main/resources/screens/live/IEDET_SEND_INPUT.json
```diff
@@ -22,18 +24,40 @@
...
+            "label": "Partially Substantiated",
+            "value": "PartiallySubstantiated"
+          },
+          {
+            "label": "Withdrawn",
+            "value": "Withdrawn"
+          },
+          {
+            "label": "Other",
+            "value": "other"
```
JosephRogerson, line 53:
```suggestion
            "value": "Other"
```

JosephRogerson (approved):

PR #997, opened 2023-04-14
HOCS-5673: Updated transfer to CCH screens for COMP and COMP2

AlexW013:
Made required content changes to 'Transfer to CCH' screens for COMP and COMP2. 

These screens exist for Service, Ex-Gratia and Minor Misconduct workflows across COMP and COMP2, however, they share the same json file.

There was a change to remove the 'Transfer to' field however this value is still required to be passed on in the workflow so the component has been changed to a hidden component with default value "CCH".

### DIFF
--- a/src/main/resources/screens/live/COMP_SERVICE_TRIAGE_TRANSFER.json
+++ b/src/main/resources/screens/live/COMP_SERVICE_TRIAGE_TRANSFER.json
@@ -1,7 +1,7 @@
 {
   "props": {},
-  "title": "Triage Transfer Case",
-  "defaultActionLabel": "Continue",
+  "title": "Transfer to CCH",
+  "defaultActionLabel": "Submit",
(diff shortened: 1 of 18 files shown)


src/main/resources/screens/live/COMP_SERVICE_TRIAGE_TRANSFER.json
```diff
@@ -10,23 +10,7 @@
       "props": {},
       "component": "text-area",
       "name": "CaseNote_TriageTransfer",
-      "label": "Enter reason for transfer"
-    },
-    {
```
LiamMacP, line 15: How does this impact old cases on the old version of the workflow that will no longer have this variable set?

LiamMacP (approved):
Approved - just needs the workflow file removing from the changes.

PR #999, opened 2023-04-17
HOCS-6544: BF2 PSU Comp type set back to Service on recategorise

AlexW013:
Updates complaint type to 'Service' if a BF2 case is rejected from PSU. This has to be done separately for the case data blob and the camunda variable CompType due to a suspected issue with camunda.

I have tested that the camunda variable is being successfully updated once the case is back with BF2:
Screenshot 2023-04-18 at 16 39 02

Also included a fix for an issue where the camunda variable CompType was not being correctly set to SeriousMisconduct upon escalation to PSU for BF, COMP and COMP2


src/main/resources/processes/BF2/BF2.bpmn
```diff
@@ -230,6 +230,7 @@
         <camunda:out source="ReturnCase" target="ReturnCase" />
         <camunda:in sourceExpression="${execution.processBusinessKey}" target="CaseUUID" />
         <camunda:in source="SeriousMisconduct" target="CompType" />
+        <camunda:in source="Service" target="CompType" />
```
LiamMacP, line 233: How does this work with the line above? 232 sets CompType to SeriousMisconduct and then this sets to Service. This would override L232?

src/main/resources/processes/PSU/PSU_BF2_COMPLAINT.bpmn
```diff
@@ -52,32 +52,41 @@
       <bpmn:outgoing>Flow_0f7331e</bpmn:outgoing>
     </bpmn:serviceTask>
     <bpmn:sequenceFlow id="Flow_0f7331e" sourceRef="Service_UpdateCompType" targetRef="CallActivity_PSU" />
+    <bpmn:serviceTask id="Activity_105d17w" name="Update complaint type" camunda:expression="${bpmnService.updateCaseValue(execution.getVariable(&#34;CaseUUID&#34;), execution.processBusinessKey, &#34;CompType&#34;,  &#34;Service&#34;)}">
```
LiamMacP, line 55: Can we have a useful ID on this task?

jeff-horton-ho-sas (approved):

PR #1000, opened 2023-04-18
HOCS-6550: support live cases ministerial sign off

LiamMacP:
HOCS-6199 broke backwards compatibility with existing cases on old live process versions. This change aims to re-add backwards compatibility into this flow and screens to allow these cases to progress.

The caveat is that cases created since the last release will have a small timeframe where `ACCEPT` won't work. As this option is for existing cases, they will be able to reselect the relevant new option.

### DIFF
--- a/src/main/resources/screens/live/APPROVE_MINISTER_SIGN_OFF.json
+++ b/src/main/resources/screens/live/APPROVE_MINISTER_SIGN_OFF.json
@@ -10,6 +10,10 @@
       "props": {
         "choices": [
           {
+            "label": "Yes (Legacy)",
+            "value": "ACCEPT"
+          },
+          {
             "label": "Yes - with a wet signature",
             "value": "ACCEPT_WET"
           },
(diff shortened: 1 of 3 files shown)


src/main/resources/screens/live/APPROVE_MINISTER_SIGN_OFF.json
```diff
@@ -9,6 +9,10 @@
       ],
       "props": {
         "choices": [
+          {
+            "label": "Yes (Legacy)",
```
jeff-horton-ho-sas, line 13: Does this need adding to the accordion config in hocs-casework?

jeff-horton-ho-sas (approved):

src/main/resources/screens/live/APPROVE_MINISTER_SIGN_OFF.json
LiamMacP: It will do - I'll follow up with that shortly - just want to get this in.

PR #1001, opened 2023-04-19
HOCS-6550: support live cases ministerial sign off

LiamMacP:
HOCS-6199 broke backwards compatibility with existing cases on old live process versions. This change aims to re-add backwards compatibility into this flow and screens to allow these cases to progress.

The caveat is that cases created since the last release will have a small timeframe where `ACCEPT` won't work. As this option is for existing cases, they will be able to reselect the relevant new option.


LiamMacP:
Hotfix release completed.

PR #1011, opened 2023-05-04
HOCS-6603 CCT Triage Accept - Transfer to CCH

robertdeniszczyc2:
Changes value of the transfer to CCH to No to restore backwards compatibility on existing cases


LiamMacP:
Closed as hotfix applied.

PR #1012, opened 2023-05-05
Fix: Update COMP triage accept screens, reject option

robertdeniszczyc2:
The reject to CCH option had been set with the value CCH This broke existing cases as the old workflows expected a Yes/No value This change updates the CCH value to be no, so it becomes backwards compatible


src/main/resources/processes/COMP/COMP_EXGRATIA_TRIAGE.bpmn
```diff
@@ -163,7 +163,7 @@
       <bpmn:conditionExpression xsi:type="bpmn:tFormalExpression">${CctTriageAccept == "Yes"}</bpmn:conditionExpression>
     </bpmn:sequenceFlow>
     <bpmn:sequenceFlow id="SequenceFlow_0yakx66" sourceRef="ExclusiveGateway_1hbd7xu" targetRef="Screen_Transfer">
-      <bpmn:conditionExpression xsi:type="bpmn:tFormalExpression">${CctTriageAccept == "No" || CctTriageAccept == "CCH"}</bpmn:conditionExpression>
+      <bpmn:conditionExpression xsi:type="bpmn:tFormalExpression">${CctTriageAccept == "No"}</bpmn:conditionExpression>
```
jeff-horton-ho-sas, line 166: Should CCH be left in so we can eventually switch back to CCH once old cases are cleared out?

jeff-horton-ho-sas (approved):
One comment, but happy to accept as is if you disagree.

Do we also need to update the label for this value in [hocs-casework](https://github.com/UKHomeOffice/hocs-casework/blob/main/src/main/resources/config/details/stages/COMP.json#L972)?

robertdeniszczyc2:
Converted PR to Draft while we confirm impact of outstanding questions

LiamMacP:
Moved to Ready as business confirmed they are happy with the change - HOCS-6603

LiamMacP (approved):

PR #1046, opened 2023-08-31
HOCS-6728: Give users ability to move cases back from offer stage

JosephRogerson:


jeff-horton-ho-sas (approved):
One typo to correct otherwise looks good

PR #1111, opened 2025-07-03
[HOCS-7026] Add confirmation of sending reply to DCU for TRO cases

jeff-horton-ho-sas:
https://collaboration.homeoffice.gov.uk/jira/browse/HOCS-7026


farrukhishaq (approved):
looks like its a schema consumed by some service, am I right to assume that it will summon some buttons on UI?

PR #1112, opened 2026-05-20
end-to-end-target-on-home-sec-cases-from-10-to-20

farrukhishaq:
That BPMN task is not just metadata; it actively **overrides stage deadlines** in casework via `bpmnService.updateDeadlineForStages(...)`.  
So after switching Home Sec case deadline to 20 days (`updateDeadlineDays(...,20)`), keeping old stage values (`7/9/10`) would cause mismatches.

Why the change was needed:
- If case deadline = 20 but stage deadlines still end at 10, users would see stage targets inconsistent with the new E2E policy.
- Workflow behavior/tests in `DCU_MIN_Markup` and `DCU_MIN_PRIVATE_OFFICE` assert both case and stage deadline updates, so they must align.
- Existing “other Private Office” path already used a 20-day profile with stage milestones `10,10,19,19,20,20,20,20`; reusing that keeps behavior consistent across PO types.

Why those exact values:
- `DCU_MIN_INITIAL_DRAFT`, `DCU_MIN_QA_RESPONSE` → `10` (midpoint milestones)
- `DCU_MIN_PRIVATE_OFFICE`, `DCU_MIN_MINISTER_SIGN_OFF` → `19` (warning/near-deadline stages)
- final stages (`TRANSFER_CONFIRMATION`, `NO_REPLY_NEEDED_CONFIRMATION`, `DISPATCH`, `COPY_NUMBER_TEN`) → `20` (full E2E target)

So in short: this edit was required because changing only the top-level case SLA would have left Home Sec stage SLAs on the old 10-day schedule, creating contradictory deadlines.

### DIFF
--- a/src/main/resources/processes/DCU/DCU_MIN_MARKUP.bpmn
+++ b/src/main/resources/processes/DCU/DCU_MIN_MARKUP.bpmn
@@ -331,7 +331,7 @@
       <bpmn:outgoing>SequenceFlow_1tvrupz</bpmn:outgoing>
       <bpmn:outgoing>SequenceFlow_0h5w2hd</bpmn:outgoing>
     </bpmn:exclusiveGateway>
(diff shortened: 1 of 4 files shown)


mohomeoffice (approved):
looks okay to me although I wasnt sure about values

jeff-horton-ho-sas (approved):

